PR #9, opened 2023-02-04
Specify which service to pull and update.

maanoobh:
Currently if you have multiple services that have updates in one docker-compose.yml file, the script updates all the services in the file which might not be what the user chose.

This just specifies the service in the commands.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -135,9 +135,10 @@ if [ -n "$GotUpdates" ] ; then
   if [ "$UpdYes" != "${UpdYes#[Yy]}" ] ; then
     for i in "${SelectedUpdates[@]}"
     do 
-      ContPath=$(docker inspect "$i" --format '{{ index .Config.Labels "com.docker.compose.project.config_files"}}')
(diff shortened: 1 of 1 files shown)


mag37:
Great find! 
I never considered this as the usual process is `docker compose pull && docker compose up -d` for most people (including myself).
But it's not clear to the user - as you could select one service in a stack but not the other from the list so this makes sense.

Though when testing your suggestions I find the first suggestion should suffice as the service name should always be the same as `$i` . Or have you seen a different outcome?

maanoobh:
I thought the same, but if there is no "container_name" in the compose file to explicitly set the name, then Docker will append the parent directory name (project name) to the service name, which will make the command fail.

[https://docs.docker.com/compose/reference/#use--p-to-specify-a-project-name](url)

mag37:
You're right! Good call! 
I'll merge this to main in a bit, doing some work on `testing` branch which I'll merge after.
Thank you for pointing out the error and finding a solution! Good work.

PR #12, opened 2023-02-09
Update dockcheck.sh

nnmalex:
Current implementation is looking for the `docker-compose.yml' in the current directory. I've added the path handling, so it takes the full path to the file location in consideration.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-VERSION="v0.1.4"
+VERSION="v0.1.5"
 Github="https://github.com/mag37/dockcheck"
 
 ### Check if there's a new release of the script:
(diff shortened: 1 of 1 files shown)


mag37:
>Current implementation is looking for the `docker-compose.yml' in the current directory. I've added the path handling, so it takes the full path to the file location in consideration.

I'm afraid this doesnt seem correct. 

The original command    
`docker inspect freshrss --format '{{ index .Config.Labels "com.docker.compose.project.config_files" }}'`   
is looking up the absolute full path of the compose-file for the service itself.
While `com.docker.compose.project.working_dir` is returning the parent directory of the compose-file. 

So with the suggestion you'd get the error    
`stat /home/User/dockers/dokuwiki/home/User/dockers/dokuwiki/docker-compose.yml: no such file or directory`    
as it takes the full path of the project followed by absolute full path of the compose-file.

I'll close this later if nothing else comes up - I've tested the changes and they're counter productive.

nnmalex:
I've just checked on multiple Debian based hosts and they all have the same label structure with `config_files` not having the full path. E.g.:
```
 "Labels": {
                "com.docker.compose.config-hash": "%hash%",
                "com.docker.compose.container-number": "1",
                "com.docker.compose.oneoff": "False",
                "com.docker.compose.project": "portainer",
                "com.docker.compose.project.config_files": "docker-compose.yaml",
                "com.docker.compose.project.working_dir": "/home/dockerhost/portainer",
                "com.docker.compose.service": "portainer",
                "com.docker.compose.version": "1.29.2"
            }
```
whereas if I test it on a Mac I get the full path in the `config_files` label.

Let me re-write this to check if the `config_files` label contains slashes and take this into account.

mag37:
That's odd, good find! I wonder what causes the difference.
I run different versions of Fedora (and Fedora Server) and also a couple of debian-based machines and I've tested it on all of them without that issue. Got `"com.docker.compose.version": "2.15.1"` on the current machine I'm testing this on. If that makes any difference.

I'll check my debian-machines again later today. Either way, would be nice to take it into account.

nnmalex:
> I wonder what causes the difference.

I've tried figuring out what the differences were in my setup, but there's nothing apparent on the first sight.
Checked all my hosts and see different behaviour with the latest version of compose available from the repos.

mag37:
>Checked all my hosts and see different behaviour with the latest version of compose available from the repos.

Well the leap is pretty big from 1.29 to 2.15! So do you get full path on the ones with higher version but only filename on the old?

I'll merge, it's a good change for compatability. 
Just interested in what's causing it.

nnmalex:
> So do you get full path on the ones with higher version but only filename on the old?

This seems to be the case, yes.

PR #16, opened 2023-02-11
Execute up command from the compose file folder

F0x06:
Hi, 

First of all, thanks for this really cool script.

I'm using a special structure for my containers, in the case of relative paths inside docker-compose.yml file, things can break, for example:

```
├── docker-compose.yml
└── storage
    ├── vol1
    └── vol2
```

```yml
services:
  some_app:
    volumes:
      - $PWD/storage/vol1:/vol1
      - $PWD/storage/vol2:/vol2
```

Hope this helps,

Regards,
F0x

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -165,6 +165,7 @@ if [ -n "$GotUpdates" ] ; then
       else
         ComposeFile="$ContPath/$ContConfigFile"
       fi
+      cd "$(dirname "${ComposeFile}")"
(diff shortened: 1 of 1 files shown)


mag37:
Thank you, glad you like it.
And thank you for bringing this to my attention. I've always been running with either absolute paths or `./data:data` so never saw this in action! Did some testing and you're right, there should be a cd in there.

I'll merge this now and tomorrow I'll merge some more fixes prior to a new release. Just need to do some more testing.

PR #23, opened 2023-02-15
Fix update check

nnmalex:
With longer first line the version check doesn't work correctly any longer.


mag37:
Thank you! Merging!

PR #41, opened 2023-08-28
Update dockcheck.sh to include stopped containers

elchanly:
as suggested in reddit, this is proposed to include stopped containers take into account im not skilled in this programming language but i wanted to contribute.


mag37:
Thanks a lot for the input! I did some modifications, hope you dont mind. Merging..

PR #46, opened 2023-12-08
Execute script in the script's path

Impact123:
`cd`s into the script's directory. The goal was to be able to run the script from anywhere without having the `regctl` binary downloaded to where it's called and this was the simplest way. It's often a good idea in other cases too.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+cd "$(dirname "$0")"
 VERSION="v0.2.5"
 ### ChangeNotes: Added an -s option to include stopped contianers in the check.
 Github="https://github.com/mag37/dockcheck"


mag37:
Added a recent fix that checked if `regctl` existed in the scripts directory, see [commit](https://github.com/mag37/dockcheck/commit/3d1c424d234aee6a303fd193191cab9c90e151f5).    
I think you're right that its sensible to cd to the scripts directory in many cases.   

But some adds the script as a "binary" in their PATH eg `/usr/bin/dockcheck` or `~/bin/dockcheck` (removing the .sh extension) to be able to just write `$ dockcheck -yp` for example, no matter their current directory.    

Then a cd might be odd and unnecessary. I'll leave the PR open and consider adding it anyway - though the recent commit should solve your regctl-issue.

Impact123:
Assuming I install it like this
```bash
root@debian:~# wget -O /usr/local/bin/dockcheck https://raw.githubusercontent.com/mag37/dockcheck/main/dockcheck.sh^C
root@debian:~# chmod +x /usr/local/bin/dockcheck
```
The issue is still the same
```bash
root@debian:~# dockcheck
Required dependency 'regctl' missing, do you want it downloaded? y/[n] y
...

No updates available, exiting.

root@debian:~# ls
regctl

root@debian:~# cd /tmp/
root@debian:/tmp# dockcheck
Required dependency 'regctl' missing, do you want it downloaded? y/[n] y
...

No updates available, exiting.

root@debian:/tmp# ls
regctl  ...
```
In my case I simply clone it into my home directory and call it like `~/dockcheck/dockcheck.sh`.
Usually from my home directory but above is just to illustrate my point that the `regctl` download functionality is a bit flawed.

mag37:
You've certainly made your case 😁.
And indeed the regctl check/handling is flawed, just been bolting on things instead of thinking it through completely.

I'll look into cleaning it up some! 
Your suggestion on different download destinations based on how its run seems reasonable. I'll to some testing and think of possible implications due to how people run and update the software. Maybe I should add a version-check of regctl too while I'm at it.

mag37:
I did some changes per your suggestions, sorry I didnt attribute you in the commit. I'll do it in the release if you approve it as "solved" even though I didnt add the cd bit, but rather went with changing how regctl is checked.

Impact123:
I think the suggestion in my **TLDR** is the best way to handle it now that I had more time to think about it but his fixes my original issue so I'm gonna mark it as solved. Thank you!

Issue #43, opened 2023-11-02
[FEATURE REQUEST] email notification (patch included)


yoyoma2:
This script is great. I just wanted to add email notification for my synology DSM and it works with a few lines.

[DSMemailnotification.txt](https://github.com/mag37/dockcheck/files/13232993/DSMemailnotification.txt)

It's quite specific to DSM emails however. How can I still benefit from the script's nice self update feature if email notifications don't make it in?

One line in the patch is a general fix to find regctl when running as a cron job (the script itself and regctl are not in the path but regctl is with the script. This line should make it in regardless of email notification.

mag37:
Sorry for the late reply! The point about the regctl binary check is a good one, thank you for reporting it.
The email-notification is a good suggestion, thou I wont add it for now - mainly because its pretty niche and requires external functionality (ssmtp).
If you want notifications there's the spinoff-project [palleri/DCW](https://github.com/Palleri/DCW) that handles all the big notification protocols.

I'll leave this open for now for consideration, and until I've added the regctl-fix. 
Thank you again.

yoyoma2:
Agreed it's niche to synology DSM. That's why I didn't make a hopeless pull request. I did see that brother project and it's overkill and seems complicated to set up. This is working great and I got docker update notification emails from my NAS this week.

Glad that the regctl-fix might make it in.

Idea:
Create a file called notify_DSM.sh.sample (it contains send_notification() that works with synology DSM)
If file notify.sh exists it is sourced and send_notification() is called in the right place. (no -m needed)
To use it, DSM users copy/tweak notify_DSM.sh.sample to notify.sh and the feature works.

Other future users submit pull requests adding files notify_XYZ.sh.sample for their XYZ platform.
Users can still benefit from this script's nice self-update feature.

Note:
I explain poorly. I could make a pull request showing this if it makes sense to you.

mag37:
Sorry, I completely missed your Idea+Note-section of the last comment.

It sounds like a good idea, though maybe quite narrow in usecase still. I'll look into it if I get time to spare!
I'll leave the issue open either way.

PR #51, opened 2024-01-01
add a generic sample notification and a DSM notification addon

yoyoma2:
This is a followup on Issue #43.

Only two lines are changed in the dockcheck.sh script that run a notification addon if one is present in notify.sh.

notify_generic.sh
A generic "hello world" notification addon. If someone wants to create a notification addon for another platform this is a starting point.

notify_DSM.sh
This is a working notification addon for synology DSM. The _SendMailTo_ variable should be set to the user's own email address. The file must be copied/renamed to notify.sh in order to run. A symbolic link does the trick also.

The dockcheck.sh script's update from git feature could work for addon users since notify.sh is not distributed. A "WARNING: Thirdparty addon use at your own risk" could be added to this and future addons. Now updates are a pain.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -14,6 +14,8 @@ ScriptWorkDir="$(dirname "$ScriptPath")"
 LatestRelease="$(curl -s -r 0-50 $RawUrl | sed -n "/VERSION/s/VERSION=//p" | tr -d '"')"
 LatestChanges="$(curl -s -r 0-200 $RawUrl | sed -n "/ChangeNotes/s/### ChangeNotes: //p")"
 
+[ -s $ScriptWorkDir/notify.sh ] && source $ScriptWorkDir/notify.sh
+
 ### Help Function:
 Help() {
   echo "Syntax:     dockcheck.sh [OPTION] [part of name to filter]" 
@@ -232,6 +234,7 @@ fi
 if [[ -n ${GotUpdates[*]} ]] ; then 
    printf "\n%bContainers with updates available:%b\n" "$c_yellow" "$c_reset"
    [[ -z "$AutoUp" ]] && options || printf "%s\n" "${GotUpdates[@]}"
+   [[ $(type -t send_notification) == function ]] && send_notification ${GotUpdates[@]}
 fi
 
 ### Optionally get updates if there's any 
--- a/notify_DSM.sh
+++ b/notify_DSM.sh
@@ -15,15 +15,15 @@ ssmtp $SendMailTo << __EOF
 From: "$FromHost" <$SendMailTo>
 date:$(date -R)
 To: <$SendMailTo>
-Subject: [diskstation] Some docker packages need to be updated
+Subject: [diskstation] Some docker containers need to be updated
 Content-Type: text/plain; charset=UTF-8; format=flowed
 Content-Transfer-Encoding: 7bit
 
-The following docker packages on $FromHost need to be updated:
+The following docker containers on $FromHost need to be updated:
 
-"$UpdToString"
+$UpdToString
 
-From $FromHost
+ From $FromHost
 
 __EOF
 }
--- a/notify_DSM.sh
+++ b/notify_DSM.sh
@@ -1,7 +1,8 @@
 # copy/rename this file to notify.sh to enable email notifications on synology DSM
 
 send_notification() {
-
+Updates=("$@")
+UpdToString=$( printf "%s\n" "${Updates[@]}" )
 # change this to your usual destination for synology DSM notification emails
 SendMailTo=me@mydomain.com
 FromHost=$(hostname)
@@ -18,8 +19,9 @@ Content-Transfer-Encoding: 7bit
 
 The following docker packages on $FromHost need to be updated:
 
-$@
+"$UpdToString"
 
- From $FromHost
+From $FromHost
+
 __EOF
 }
--- a/notify_DSM.sh
+++ b/notify_DSM.sh
@@ -1,4 +1,6 @@
-# copy/rename this file to notify.sh to enable email notifications on synology DSM
+### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
+# Copy/rename this file to notify.sh to enable email notifications on synology DSM
+# Modify to your liking - changing SendMailTo and Subject and content.
 
 send_notification() {
 Updates=("$@")
--- a/notify_DSM.sh
+++ b/notify_DSM.sh
@@ -1,0 +1,25 @@
+# copy/rename this file to notify.sh to enable email notifications on synology DSM
+
+send_notification() {
+
+# change this to your usual destination for synology DSM notification emails
+SendMailTo=me@mydomain.com
+FromHost=$(hostname)
+
+printf "\nSending email notification\n"
+
+ssmtp $SendMailTo << __EOF
+From: "$FromHost" <$SendMailTo>
+date:$(date -R)
+To: <$SendMailTo>
+Subject: [diskstation] Some docker packages need to be updated
+Content-Type: text/plain; charset=UTF-8; format=flowed
+Content-Transfer-Encoding: 7bit
+
+The following docker packages on $FromHost need to be updated:
+
+$@
+
+ From $FromHost
+__EOF
+}
--- a/notify_generic.sh
+++ b/notify_generic.sh
@@ -2,11 +2,11 @@
 # generic sample, the "Hello World" of notification addons
 
 send_notification() {
+  Updates=("$@")
+  UpdToString=$( printf "%s\n" "${Updates[@]}" )
+  FromHost=$(hostname)
 
-FromHost=$(hostname)
-
-# platform specific notification code would go here
-printf "\n%bGeneric notification addon:%b" "$c_green" "$c_reset"
-printf "\nThe following docker packages on $FromHost need to be updated:\n$@\n"
-
+  # platform specific notification code would go here
+  printf "\n%bGeneric notification addon:%b" "$c_green" "$c_reset"
+  printf "\nThe following docker packages on %s need to be updated:\n%s\n" "$FromHost" "$UpdToString"
 }
--- a/notify_generic.sh
+++ b/notify_generic.sh
@@ -1,0 +1,12 @@
+# copy/rename this file to notify.sh to enable email/text notifications
+# generic sample, the "Hello World" of notification addons
+
+send_notification() {
+
+FromHost=$(hostname)
+
+# platform specific notification code would go here
+printf "\n%bGeneric notification addon:%b" "$c_green" "$c_reset"
+printf "\nThe following docker packages on $FromHost need to be updated:\n$@\n"
+
+}
--- a/notify_generic.sh
+++ b/notify_generic.sh
@@ -9,5 +9,5 @@ send_notification() {
 
   # platform specific notification code would go here
   printf "\n%bGeneric notification addon:%b" "$c_green" "$c_reset"
-  printf "\nThe following docker packages on %s need to be updated:\n%s\n" "$FromHost" "$UpdToString"
+  printf "\nThe following docker containers on %s need to be updated:\n%s\n" "$FromHost" "$UpdToString"
 }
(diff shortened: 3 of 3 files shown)


mag37:
This looks very promising! Thank you for the work. What do you mean by **Now updates are a pain.** ?
I'll read through and test it tomorrow - and merge it if nothing comes up :)

mag37:
The `notify_generic.sh` wouldnt list more than the first item of the array due to expanding without index. It's most likely the same with the DSM-version which I didn't test.
I changed the `notify_generic.sh` to this to make it work:
```bash
# copy/rename this file to notify.sh to enable email/text notifications
# generic sample, the "Hello World" of notification addons

send_notification() {
Updates=("$@")
FromHost=$(hostname)

# platform specific notification code would go here
printf "\n%bGeneric notification addon:%b" "$c_green" "$c_reset"
printf "\nThe following docker packages on %s need to be updated:\n%s \n" "$FromHost" "${Updates[*]}"

}
```

Also I'd quote the variable when calling the function like this:
`[[ $(type -t send_notification) == function ]] && send_notification "${GotUpdates[@]}"`

Would you mind doing those changes and I'll merge? :smile:

yoyoma2:
> Would you mind doing those changes and I'll merge? 😄

Good catch! I only use a couple docker containers on my NAS and only one at a time has needed updates since I've been using my tweaked dockcheck. Yes the DSM version has the same bug as the generic.

Is there a small container that gets nightly builds to use for testing? I guess I could run an older version of anything to get multiple containers needing updates. I have one requiring update now. I'm a docker noob...

**Happy to fix it myself** as long as I can test multiple pending updates or if you merge your fixed version and close my buggy pull request that's OK too.

mag37:
Yeah, easy to miss no worries :)

Hmm. Not that I know of. I've got a bunch of containers on a few machines, so I'll just always have some to update. 
But I've got a lab environment for testing breaking stuff - there I'll just `docker image rm -f b4bz/homer`, that will throw some errors (due to not finding the image to compare with) and its not good for the container - but it'll show up as needing updates.

I think I can make changes to your fork and then merge it, if that's okay.

yoyoma2:
Perfect. I only have containers on the NAS and I can't break stuff there. I'll test any new version you provide on my DSM NAS.

A disclaimer at the top of the DSM and any future notification addons people might provide might be a good idea. You only need to test/maintain dockcheck itself.

yoyoma2:
This morning I installed pihole nightly and tonight it already needs an update so I can now test two outdated containers.

The email I get from the NAS correctly shows both outdated packages however on a single line so not pretty.
The generic addon only prints the first one as you said.
I also got warnings from diff about no newline at end of file.
I'll take a look tomorrow.

yoyoma2:
The generic addon works perfectly.
The DSM addon displays on multiple lines now but the quotes appeared in the email.
Also the space I had before the word _From_ is needed otherwise Thunderbird thinks it's a keyword and displays a vertical bar before it. The space does not appear in the email. This email looks exactly like a DSM email and I had to view source on theirs to figure out the space before _From_ trick.

Should the word _package_ be replaced by container everywhere in both addons? I copied from DSM emails but you're the docker expert.

```
diff --git a/notify_DSM.sh b/notify_DSM.sh
index e5283e1..de89f5e 100644
--- a/notify_DSM.sh
+++ b/notify_DSM.sh
@@ -21,9 +21,9 @@ Content-Transfer-Encoding: 7bit

 The following docker packages on $FromHost need to be updated:

-"$UpdToString"
+$UpdToString

-From $FromHost
+ From $FromHost

 __EOF
 }

```
I'm looking forward to easily upgrading to a new dockcheck version and having the email notifications still work. :-)

I'll try to promote dockcheck to synology DSM users. Just a simple entry in the task scheduler. Such an elegant solution your dockcheck.

mag37:
Oh yes sorry about the space! Didn't realize I removed it.

Your edit looks good :) commit that if you havnt already.

I'm no docker expert 😅 nor bash developer, just guessing and trying. But yeah, I think **container** is probably a better wording.

Nice work! Does it seems to work as expected with your changes now or anything else we need to fix?

yoyoma2:
I committed the changes including rewording to **container**. It works perfectly on my NAS. Ready to merge IMHO.

To Do:
Since the NAS is already configured to know the recipient's email address it would be nice if SendMailTo defaulted to that value but I didn't find where this setting is stored so manually setting it is required for now.

This works on a vanilla DSM installation. If someone installs some mail related DSM packages they may want to add support for other email sending tools in addition to the default ssmtp.

mag37:
Wonderful, thank you for your work! I'll merge it now.

If you work further on it and change *SendMailTo*, just let me know or create another PR!

yoyoma2:
I just looked at the GUI in DSM for email notifications (attached). The "Sender email" field has in "i" and the tooltip says "If blank, the first recipient's email address that you specified below will be used.". Mine is blank so that's what I receive. The addon could do the same? If left blank From does what it does now?

The addon email is identical to DSM emails except I forgot the host in the subject. Could you add it at the same time?
Subject: [diskstation] Some docker containers **on $FromHost** need to be updated

Yes I was wondering how to update the addon if it has changes. We don't want user modifications to be clobbered. Single file sounds easy to distribute but tweaking the From, To etc might get harder if many addons get added.
DSM_Mail1
DSM_Mail2

mag37:
Just did some testing on my `ssmtp` setup, and leaving it empty break things.
If I understand the DSM-wording, it setting the "From"-field to the same as the first recipient, so that'd be equal to setting the variables to:
```
SendMailFrom=me@mydomain.com
SendMailTo=me@mydomain.com
```
But having the variables To and From, opens the option to do:
```
SendMailFrom=me@mydomain.com
SendMailTo=you@yourdomain.com
```

Wouldn't it work equally good to have both the variables and just let the user decide if you'd like same or different?
Thank you for the screenshots btw!

mag37:
>Yes I was wondering how to update the addon if it has changes. We don't want user modifications to be clobbered. Single file sounds easy to distribute but tweaking the From, To etc might get harder if many addons get added.

What I was thinking with a single file was to keep every "addon" separated by some divider+title, like 
```
###### Email Template ######
send_notification() {
---addonspecific code here --
}

###### Other Template ######
send_notification() {
---addonspecific code here --
}
```

I dont know what's cleaner for the user - either one big file to copy-paste the wanted snippet from or specific files in a directory to just copy+edit.

And about not breaking the user modifications - maybe you'll be required to use .gitignore for the `notify.sh`.. I'll have to read up if there's other options when ppl clone the project.

yoyoma2:
The single file would be great if it was automatically updated alongside dockcheck.sh when a new release is detected. Doing a diff to see the changes between the user's notify.sh and the templates file will be tough after each update.

What I mean to behave like DSM is to default to use the SendMailTo if SendMailFrom is null or unset. By default SendMailFrom could be commented out:

#SendMailFrom=me@mydomain.com
SendMailTo=you@yourdomain.com

echo SendMailFrom=$SendMailFrom
echo SendMailTo=$SendMailTo

echo Using SendMailFrom ${SendMailFrom:-$SendMailTo}

For anyone using ssmtp your change will be useful not just on DSM. Nice!

mag37:
I just created a branch where I'm doing some additions and changes, check it out if you like:
https://github.com/mag37/dockcheck/tree/notify_fixes

mag37:
I will think a bit more on single file VS a directory of templates.
The users will still need to choose their template, copy it to the same directory as their `dockcheck.sh` and do their modifications to make it work on their system (eg. setting the To+From etc). 
And when they're done, their `notify.sh` wont be touched.

I'll just have two separate templates for DSM and ssmtp for now :) 

If you want you could to try the ssmtp-template I made and see if it works good for DSM (if you just comment out the From as is).

Again, thank you for the cooperation and initiative to make these changes!

mag37:
Continued on the branch [notify_fixes](https://github.com/mag37/dockcheck/tree/notify_fixes) tonight.

Added a new option `-i` to trigger the notification - realized that people might wanna have the notification sometimes and run the regular update etc without spamming notifications other times :)

I'll do some more testing and take any input you have about the templates before pushing to main.

yoyoma2:
I'll look at the notify branch and give feedback after a night's sleep. Will try your ssmtp flavor on DSM and see how that goes. I like the `-i` idea. I already spammed myself a few times.

yoyoma2:
I tried the DSM and the ssntp versions and they both work on DSM. Yay! The two could be in the same ssntp addon file using variables set in a switch/case according to a flavor (flavour in UK). At the top:

\# set flavor to dockcheck, DSM, custom
flavor=dockcheck

case dockcheck
  set variables
case DSM
  set variables
case custom
  set variables

same email sending code using variables.

Could you please add the $FromHost in the subject line on DSM as shown 4 lines above the screenshots? The ssntp version's subject has the host.

I like the gitignore for notify.sh

I like the `-i`

I like the apprise idea. Work in progress?

mag37:
Wonderful! Thank you for testing and giving feedback.

What did you need changing between DSM and sSMTP? (what differs in the "flavour" that is?)
I'll add the $FromHost to subject either way!

The apprise-template works as is, people have their own apprise-setup (either native, bare metal installation or the dockerized API version) and can trigger their notifications with the template.
Apprise is pretty cool, you can add a plethora of notifications (discord, telegram, matrix, email etc) and then trigger them all or selective based on tags.

mag37:
Also: 
Did the ssmtp-version work even though it lacked the (space)$FromHost ?
I figured the $FromHost is already referenced in the subject and first line, so didnt need it a third time.

yoyoma2:
Pulled and re-tested both on my NAS. Set SendMailTo on DSM and both SendMailTo and SendMailFrom on ssmtp to the same thing. Both work great! If you had a Signature variable (empty for ssmtp) like the new SubjectTag then both versions are identical.

Just make sure that if SendMailFrom's assignement is commented out it defaults to SendMailTo's value later.

Oh and thanks for the tweak to the DSM subject.

mag37:
I'm not entirely sure I understand why you'd need to have logic around assigning the SendMailFrom variable, instead of just having the user set both variables (SendMailFrom + SendMailTo) to the same address, if that's what's required by DSM.

And about the Signature - is it required, I mean does it break if you don't have `(space)From $FromHost`?
Like when you tested the ssmtp-template, did you have to modify it for it to work?

I'm just thinking it's templates, a user can modify it however they like, if it doesn't mean breakage.
Though if I misunderstand something I'll stop hounding about it ^^

For now I'll just keep both templates as is! Also clearer for the user that DSM-features are supported, good advertising.

Thank you again for the work you've put in.

yoyoma2:
DSM is all GUI based and attracts people who like things to be easy. Editing 1 thing is better than editing 2 things. If the already configured DSM destination email notification address could be read and automatically used and it became editing 0 things that would be the best. That was my thinking. You can see that DSM mindset in the screenshot above where "Sender email" is normally left blank.

It still works without any signature. I just made everything identical to other notification emails from DSM. If synology someday adds a checkbox "Send notifications for docker container updates" they will likely look exactly like these notifications from dockcheck.

By default DSM is 100% GUI. You have to manually enable ssh access and user home directories which I did immediately. :-)

Two files indeed is good advertising. Once things are working these files will probably rarely change so maintenance of very similar files isn't a big concern.

Very pleasant cooperating with you. Congrats on a nice tool!

mag37:
Ah yes, now I understand and it sounds sensible indeed. Keep it clean and simple to fit the userbase of the product.
I'll keep them separated then - seems like the best choice, both to simplify and to make it clear that it's made to support the DSM platform.

Oh 100% GUI sounds like my nightmare but I know it has it's usecases. SSH ftw.

Indeed, I think they're pretty much `set and forget` right now, maybe some minor tweaks will surface - but still easy to do.

Thank you - I would've never thought it would grow outside my circle of friends tbh. 
I actually just completed a blog post about the project tonight.
It's around the 1 year anniversary since the first commit :tada: !

If you wanna read it > https://mag37.org/posts/project_dockcheck/

yoyoma2:
Interesting read. Take a look at [this thread](https://www.synoforum.com/threads/easy-docker-container-update-notification-emails.12469/) I started on synoforum.

mag37:
Wonderful! Thank you for spreading the word.
Also - updated the readme with some info, both about the notifications and the non amd64/arm64 workaround.

PR #58, opened 2024-01-24
modified update notification

thomas-mc-work:
Slightly changed the notation of the local/latest version to simplify the visual impression. But it's a matter of taste.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -145,7 +145,7 @@ progress_bar() {
 }
 
 ### Version check & initiate self update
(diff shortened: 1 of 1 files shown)


thomas-mc-work:
Thank you!

I'm not 100 % sure, but I've noticed that the Unicode adoption is pretty well since some years. And I find those glyphs more and more i nshell scripts.

mag37:
Just tried it in the 5 different machines I got access to right now (4 different distros, 3 different prompts/shells) and it worked fine in all.
I'll merge it.

Thank you for the contribution!

PR #69, opened 2024-02-10
Added Matrix-Synapse notification Template

silverbios:
Similar to other templates like telegram, this script can send notification about avaliable Update Images to Synapse server.

### DIFF
--- a/notify_matrix.sh
+++ b/notify_matrix.sh
@@ -21,7 +21,7 @@ send_notification() {
     MatrixServer="Enter Your HomeServer URL"
     MsgBody="{\"msgtype\":\"m.text\",\"body\":\"$MessageBody\"}"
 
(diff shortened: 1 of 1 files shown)


mag37:
Superb! Thank you for the contribution!

hellblue:
how can i test if notification works for matrix from conole with notify.sh
i added my data in the file

mag37:
> how can i test if notification works for matrix from conole with notify.sh i added my data in the file

You could either just test the curl string - manually set the variables you filled in the file and set something like `Test123` as the MsgBody
```
curl -sS -o /dev/null --fail -X PUT "$MatrixServer/_matrix/client/r0/rooms/$Room_id/send/m.room.message?access_token=$AccessToken" -H 'Content-Type: application/json' -d "$MsgBody"
```

Or you could do `source notify.sh; send_notification "Test123" `

hellblue:
thank you i got it working now

PR #76, opened 2024-04-29
Update HTTP method in notify_matrix.sh

silverbios:
Change the HTTP method to POST because, in recent versions, the Synapse server only allows POST for this path.


mag37:
Thank you for maintaining the template!

PR #87, opened 2024-06-21
Create notify_gotify.sh

cerealconyogurt:
adding Gotify template


mag37:
Thank you for contributing!

PR #88, opened 2024-07-02
zero configuration DSM notification emails

yoyoma2:
Update the DSM notification template to directly use the existing synology DSM configuration. Manual editing of the DSM template is no longer required.

Shamelessly copied from the Tailscale update notifications script on SynoForum.


mag37:
Good work! Merging.

PR #102, opened 2024-10-08
fix upgly typo which prevents email notifications

thomas-mc-work:


mag37:
Good call, thank you.

yoyoma2:
The DSM template has the typo as well right?

mag37:
> The DSM template has the typo as well right?

Correct, I fixed both in the branch (I'll merge as soon as we've done some more testing).

PR #128, opened 2025-02-19
Add async checking for updates for improved performance

Thaurin:
This PR adds a command-line option `-z` that will pull images asynchronously for a *significant* performance increase.

It works by using process substitution and reading the output in a while read -r loop. This (asynchronous) input is parsed and added to the correct arrays based on the first keyword (NoUpdates/GotUpdates/GotErrors/Skip). So the end result should be the same, just async.

This is experimental for now: more testing is needed. If this is deemed stable, it could be made the default, removing the `-z` parameter.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -346,7 +346,7 @@ else
 
   export -f check_image datecheck
   export Excludes_string="${Excludes[@]}" # Can only export scalar variables
-  export t_out regbin RepoUrl
+  export t_out regbin RepoUrl DaysOld
 
   # Asynchronously check the image-hash of every running container VS the registry
   while read -r line; do
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -46,6 +46,7 @@ c_blue="\033[0;34m"
 c_teal="\033[0;36m"
 c_reset="\033[0m"
 
+MaxAsync=32
 Timeout=10
 Stopped=""
 while getopts "aynpfrhlisvmc:e:d:t:" options; do
@@ -285,13 +286,12 @@ fi
 check_image() {
   i="$1"
   local Excludes=($Excludes_string)
-  local skip
-  for e in "${Excludes[@]}" ; do [[ "$i" == "$e" ]] && skip=1 ; done
-
-  if [[ $skip -eq 1 ]]; then
-    echo Skip $i
-    return
-  fi
+  for e in "${Excludes[@]}" ; do 
+    if [[ "$i" == "$e" ]]; then
+      echo Skip $i
+      return
+    fi
+  done
 
   local NoUpdates GotUpdates GotErrors
   ImageId=$(docker inspect "$i" --format='{{.Image}}')
@@ -320,6 +320,14 @@ export -f check_image datecheck
 export Excludes_string="${Excludes[@]}" # Can only export scalar variables
 export t_out regbin RepoUrl DaysOld
 
+# Check for POSIX xargs with -P option, fallback without async 
+if (echo "test" | xargs -P 10 >/dev/null 2>&1)  ; then
+  XargsAsync="-P $MaxAsync"
+else
+  XargsAsync=""
+  printf "%bMissing POSIX xargs, consider installing 'findutils' for asynchronous lookups.%b\n" "$c_red" "$c_reset"
+fi
+
 # Asynchronously check the image-hash of every running container VS the registry
 while read -r line; do
   ((RegCheckQue+=1))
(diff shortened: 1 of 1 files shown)


Thaurin:
Alright, `-d` fix seems to have been simple. Just forgot to export `$DaysOld`. Output with `-d` seems to be the same between old method and parallel now.

mag37:
Looks really good! Been testing it some and reading through the code. 
Seems like you've covered Excludes and DaysOld fine! And nice with the error handling too.

I'd like this to be the default when merged to main, so I've removed the none-parallel bit in my testing.

Thaurin:
> Seems like you've covered Excludes and DaysOld fine! And nice with the error handling too.

Yes, excludes and days old seem to work now. I've been using it for my updates and have seen no problems yet. It's not much of error *handling*, though, just writing out an error when unexpected output comes from a subprocess. I don't know what to do it a version check totally messes up and spits out all sorts of errors. :)

If you're fine with merging it and making it the default, I'll remove the `-z` parameter and old check before merging, or you can do it. I'm stlil not sure how to handle the number of subprocesses... 32 works fine for me, it feels almost instant, but maybe it depends on bandwidth, available memory, etc. of the user.

By the way, I *think* `dockcheck` isn't really affected by [Docker Hub limiting unauthenticated users to 10 pulls per hour](https://www.reddit.com/r/selfhosted/comments/1iup6bg/docker_hub_limiting_unauthenticated_users_to_10/), unless you have >10 updates a day? Version check isn't pulling, right?

mag37:
Have just been making some other things ready (some readme-changes mainly) to merge this to main and then create a new version+release.

I was thinking of adding a `MaxAsync=32` variable up top to have users easily modify it if wanted. Then modifying the xargs command to`xargs -P $MaxAsync -I {} bash -c 'check_image "{}"' \` . 

While testing this I started thinking of the xargs package, if BSD and Alpine ships the same functions and sadly it seems that some versions of Alpines Busybox doesnt have the `-P` option nor any alternatives.
Alpine have GNU xargs in their `findutils` though but people who don't have it will hit issues.

So I'm considering making a check so if dosnt have the `-P` flag it'll throw a warning and omit the `-P #` but keep running.

Something (not that beautiful) like 
```
if (echo "test" | xargs -P 10 >/dev/null 2>&1)  ; then 
  XargsAsync="-P $MaxAsync" 
else 
  XargsAsync=""
  printf "\n%bMissing GNU xargs, consider installing 'findutils'%b\n" "$c_red" "$c_reset"
fi
```

mag37:
@Thaurin if you agree with these changes and don't have stumbled upon any other issues I'll merge this to main tonight most likely! Great work.

Thaurin:
> I was thinking of adding a `MaxAsync=32` variable up top to have users easily modify it if wanted. Then modifying the xargs command to `xargs -P $MaxAsync -I {} bash -c 'check_image "{}"' \` .

This seems like a sensible approach. In any case, I haven't seen any negative consequenses from using 32, on the contrary! And if you're running up to 32 containers, you'll probably have the RAM and bandwidth for it, right? :)

> it seems that some versions of Alpines Busybox doesnt have the -P option nor any alternatives.

Yes, this has actually crossed my mind, since you check for everything else in the script. I don't have experience with alpine (on my todo list ;).

> if you agree with these changes and don't have stumbled upon any other issues I'll merge this to main tonight most likely! Great work.

Well, you may be right that the check for the existence of the `-P` parameter could be done more nicely, but if you can test it and it works, then do it how you feel works for now. I mean, having two code paths behind an `if-else` for would not be very good, but having to parse `xargs` output to handle to adding to the correct arrays when it's not strictly needed, isn't great either.

But if it works... can't really think of anything that is much better right now.

In any case, as for testing, it has just worked on my server so far, without problems! And the increased speed is much appreciated.

mag37:
> On the other hand, with `-P 1`, the update check looks to be working exactly like the old code, i.e. one by one, which, I assume, is the same as omitting the parameter.

Yeah I tested for this - I figured the "worst" that can happen is it's as slow as before, but it wont break. So a message about it but then continue as usual with xargs but without the max processes `-P` flag.
 
> But if it works... can't really think of anything that is much better right now.
Me neither - so it'll do for now :)

> In any case, as for testing, it has just worked on my server so far, without problems! And the increased speed is really nice.
Nice to hear! Yeah I've been doing several updates and testing with `-e` exclude (changed the code around this some) and other options to see it all seems to work.

Merging it now!

mag37:
And thanks again, epic work!

Thaurin:
No problem! It's a much more evolved script than mine was and I'll integrate it with my workflow. So if I see any more improvements that I can make, I'll get to it. ;)

PR #157, opened 2025-04-06
use notification release notes url function for cli options as well

jschuster:
The feature for adding update release note urls per container is great for quick access.

This pr adds the release notes to each line in the updatable container options when the cli is used.

Why?

If you use iTerm, for example, you can then navigate directly to the release notes by CMD+clicking on the URL. This helps a lot to decide which container should receive an update and which should wait.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,8 +144,8 @@ Use a `notify_X.sh` template file from the **notify_templates** directory, copy 
 Further additions are welcome - suggestions or PR!  
 <sub><sup>Initiated and first contributed by [yoyoma2](https://github.com/yoyoma2).</sup></sub>  
 
-### :date: Release notes addon to Notifications
-There's a function to use a lookup-file to add release note URL's to the notification message.  
+### :date: Release notes addon
+There's a function to use a lookup-file to add release note URL's to the notification message and the cli updated container options.  
 Copy the notify_templates/`urls.list` file to the script directory, it will be used automatically if it's there.   Modify it as necessary, the names of interest in the left column needs to match your container names.  
 The output of the notification will look something like this:
 ```
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -292,13 +292,15 @@ dependency_check() {
 }
 
 # Numbered List function
+# if urls.list exists add release note url per line
 options() {
   num=1
-  for i in "${GotUpdates[@]}"; do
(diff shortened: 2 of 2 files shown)


yoyoma2:
Nice feature for those who use the terminal for making the upgrade decisions. For those who make those decisions in the notification, I hope a user variable can be used to enable/disable this feature which will only clutter the menu for those not using it.

mag37:
I've been meaning to implement this - it's in my todo-list! 
Thank you for the contribution. I'll need to test some things before accepting it.

And as @yoyoma2 said:
> .. I hope a user variable can be used to enable/disable this feature which will only clutter the menu for those not using it.

I will make a variable and option-flag (running out of alphabet letters soon! :grin: need to use capital letters) to let people choose weither they want this functionality.

jschuster:
Thank you, @mag37 for the wonderful tool and your quick response!

Yes, as @yoyoma2 pointed out, an option or a way to deactivate it would be good.

Just a suggestion to get around the problem with "running out of letters":
Maybe you could reinterpret the -m as ‘Minimal’ or something (since you have already thought of hiding the progress bar with this option).. 

And use the urls.list implicitly - as with the notifications.

And then use the following logic:
- If no list available - no links anyway
- If list available, links in notifications and cli options
- If parameter m, only minimal output (no progress bar, no colors, no links)

It is certainly more flexible with another option, but over time it becomes tight and possibly a little contradictory.

mag37:
> Just a suggestion to get around the problem with "running out of letters": Maybe you could reinterpret the -m as ‘Minimal’ or something (since you have already thought of hiding the progress bar with this option)..

It's a good idea - but I think this addition might be catered for a middle ground user base.
Who don't need the minimalism (some who wrap/use dockcheck in other things) or who don't use it as unattended notifications.

But those who actually look at the CLI and interact with it. So I think it should be its own option.

mag37:
Decided to merge it to a branch I'm working on (will probably merge to main today - need to do some more testing).
Thank you for the idea and the work provided!

I did some tweaks and also added a `-I` option.

jschuster:
I can fully understand how difficult it is to find the right balance.

Thank you for your effort and, in my opinion, the good solution you have chosen.

mag37:
Thank you for contributing! 
Added cred to the release notes.

PR #164, opened 2025-04-21
fix: move new line from `read` to `printf`

hasecilu:
`\n` not escaped on `read`


mag37:
Thank you, I saw this yesterday too but havn't had time yet.

PR #169, opened 2025-04-29
Detect when there is an error recreating the container

jaaviercr:
The script did not stop when  there is an error recreating the container (in this case, docker-compose.yml was moved to other folder):
```
Now updating (1/1): nginx
latest: Pulling from library/nginx
Digest: sha256:c15da6c91de8d2f436196f3a768483ad32c258ed4e1beb3d367a27ed67253e66
Status: Image is up to date for nginx:latest
docker.io/library/nginx:latest
open /home/foo/bar/docker-compose.yml: no such file or directory

All done!
```
Added an exit for for that case.
Please note this exit stops the loop, consider skipping the container, continue with the rest, and exit with a 1.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -517,7 +517,7 @@ if [[ -n "${GotUpdates:-}" ]]; then
       if [[ "$ContRestartStack" == true ]] || [[ "$ForceRestartStacks" == true ]]; then
         ${DockerBin} ${CompleteConfs} stop; ${DockerBin} ${CompleteConfs} ${ContEnvs} up -d
       else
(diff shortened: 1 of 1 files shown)


mag37:
Thank you - good point.
Just to be clear - the compose was moved prior to this (by you or something else)?
So it wasnt `dockcheck.sh` that had something to do with it being **moved**.

jaaviercr:
Yes, it was "moved" by me (actually I renamed a folder in the path and then this error happened), `dockcheck.sh` did not move the file 😊

mag37:
Wonderful, thank you for your contribution! Merged to main now.
I'll look into some better error handling in the future .. it's long due ^^

PR #175, opened 2025-05-06
Update notify_ntfy-sh.sh

op4lat:
Sure thing! 
Bumping version


mag37:
Ah I figured you'd change the open PR with the edit :) No problem though, I merged yours and then bumped it, I'll close this one.
Thank you for contributing!

PR #180, opened 2025-05-16
fix: prevent adding urls when -I is not passed

vorezal:
Address an issue where releasenotes is run in notify.sh when -I is not passed but urls.list is present.

Pass Updates[@] to send_notification to eliminate unnecessary calls to releasenotes in all cases.

My first PR here, so feedback is very welcome. I have made several other modifications of varying size and complexity for my own use, so I'll likely be submitting more shortly. If anything is out of line or against the direction you want for the project, reject away.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -481,7 +481,7 @@ if [[ -n ${GotUpdates[*]:-} ]]; then
   printf "\n%bContainers with updates available:%b\n" "$c_yellow" "$c_reset"
   if [[ -s "$ScriptWorkDir/urls.list" ]] && [[ "$PrintReleaseURL" == true ]]; then releasenotes; else Updates=("${GotUpdates[@]}"); fi
   [[ "$AutoMode" == false ]] && list_options || printf "%s\n" "${Updates[@]}"
-  [[ "$Notify" == true ]] && { type -t send_notification &>/dev/null && send_notification "${GotUpdates[@]}" || printf "\nCould not source notification function.\n"; }
+  [[ "$Notify" == true ]] && { type -t send_notification &>/dev/null && send_notification "${Updates[@]}" || printf "\nCould not source notification function.\n"; }
 fi
 
 # Optionally get updates if there's any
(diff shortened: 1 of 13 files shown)


mag37:
Thank you, I'll need to take this into some consideration though.

The line of thought (and development) was that you "activate" the urls in the notifications by just moving the file. If it's present, it'll be included.
And the `-I` option was specifically to *additionally* show the urls in the CLI.

vorezal:
Oh I see! I misunderstood the intention obviously. Now that you say that, it makes sense. Especially if you want notifications to go out with URLs, but still need to be able to view or even parse the CLI output without them. Perhaps process the CLI output with another tool, or just intentionally keep dockcheck.sh CLI arguments logically separate from notify.sh functions.

By all means, do whatever you prefer here. It just confused me when I was first setting things up, was trying to test sending notifications, and ended up switching between including and not including them several times.

mag37:
I'll ponder on this some more, maybe see if any other users got some inputs before I act.

It's still a valid point (and PR) that it might be confusing - so I'll try to think of pros and cons.

vorezal:
Based on your input, my opinion now would be not to merge this in. If anyone was using it as originally intended (with the urls.list file in place but without -I specified) and their script automatically updated, the URLs would stop appearing in notifications without warning. As for the confusion, perhaps a small change to the docs would be enough?

mag37:
You're probably right - I'll consider some changes to the docs to make the notification parts clearer - if you got any ideas I'm all ears :) type here or PR if so.

vorezal:
I think I was mainly looking at the help output and missed that it is already called out in the readme. Maybe something like this though? "Prints custom releasenote urls alongside each container with updates (requires urls.list)." -> "Prints custom releasenote urls alongside each container with updates in CLI output (requires urls.list)." in the help function and readme copy of the same? Possibly redundant though with the callout in the section on the Release notes addon though.

mag37:
> "Prints custom releasenote urls alongside each container with updates in CLI output (requires urls.list)." 

This sounds good! Might be redundant but I rather have it clear twice than causing confusion. 

I'll change this in a future fix.

mag37:
I'll close this for now - I've added it to my *todo* for the next fixup, might be your other PR ;)

PR #182, opened 2025-05-21
Refactor notifications and add helper functions

vorezal:
* Add helper functions to simplify sourcing files and executing functions if they exist
* Create notify_v2.sh wrapper script
* Simplify and consolidate notification logic within notify_v2.sh
* Support notification management via environment variables
* Move secrets to dockcheck.config

Goals:
* Do not break existing configurations. A notify.sh file already in place takes priority and must be removed for notify_v2.sh to be used.
* No major change to notifications. In my testing, migrating to the new version with an equivalent configuration resulted in identical notifications to the selected notification channel. There were minor changes to CLI output, e.g. "Sending Slack dockcheck notification" changed to "Sending dockcheck update notification - slack" and additional output was added when notifications for template updates are sent.
* Consolidate duplicated template code in the notify_v2.sh wrapper script.
* Move all sensitive and other customizable values to dockcheck.config so notify template files can be committed, sourced, and used without user modification.

I could use some help testing other notification channels I don't have set up, like Apprise and DSM, if possible. I am still working on a "snooze" extension that relies on this as a base, but I should be able to finish it up if approved. Thoughts and criticism very welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,12 @@
 ___
 ## :bell: Changelog
 
+- **v0.6.5**: Refactored notification logic. See notify_templates/notify_v2.sh for upgrade steps.
+    - Added helper functions to simplify sourcing files and executing functions if they exist.
+    - Created notify_v2.sh wrapper script.
+    - Simplified and consolidated notification logic within notify_v2.sh.
+    - Added support for notification management via environment variables.
+    - Moved notification secrets to dockcheck.config.
 - **v0.6.4**: Restructured the update process - first pulls all updates, then recreates all containers.
     - Added logic to skip update check on non-compose containers (unless `-r` option).
     - Added option `-F` to revert to `compose up -d <ContainerName>` targeting specific container and not the stack.
@@ -126,11 +132,19 @@ Alternatively create an alias where specific flags and values are set.
 Example `alias dc=dockcheck.sh -p -x 10 -t 3`.
 
 ## :loudspeaker: Notifications
-Trigger with the `-i` flag if `notify.sh` is present and configured.  
-Will send a list of containers with updates available and a notification when `dockcheck.sh` itself has an update.  
-Run it scheduled with `-ni` to only get notified when there's updates available!  
+Trigger with the `-i` flag.
+If `notify.sh` is present and configured, it will be used. Otherwise, `notify_v2.sh` will be enabled.
+Will send a list of containers with updates available and a notification when `dockcheck.sh` itself has an update.
+Run it scheduled with `-ni` to only get notified when there's updates available!
 
-Use a `notify_X.sh` template file from the **notify_templates** directory, copy it to `notify.sh` alongside the script, modify it to your needs! (notify.sh is added to .gitignore)  
+Legacy installation and configuration:
+Use a `notify_X.sh` template file from the **notify_templates** directory, copy it to `notify.sh` alongside the script, modify it to your needs! (notify.sh is added to .gitignore)
+
+V2 installation and configuration:
+Remove or rename `notify.sh` if previously configured.
+Uncomment and set the NOTIFY_CHANNELS environment variable in `dockcheck.config` to a space separated string of your desired notification channels to enable.
+Uncomment and set the environment variables related to the enabled notification channels.
+
 **Current templates:**
 - Synology [DSM](https://www.synology.com/en-global/dsm)
 - Email with [mSMTP](https://wiki.debian.org/msmtp) (or deprecated alternative [sSMTP](https://wiki.debian.org/sSMTP))
--- a/default.config
+++ b/default.config
@@ -23,3 +23,49 @@
 #PrintReleaseURL=true   # Prints custom releasenote urls alongside each container with updates (requires urls.list)`
 #PrintMarkdownURL=true    # Prints custom releasenote urls as markdown
 #OnlySpecific=true    # Only compose up the specific container, not the whole compose. (useful for master-compose structure).
+
+### Notify settings
+# All commented values are examples only. Modify as needed.
+#
+# Uncomment the line below and specify the notification channels you wish to enable in a space separated string
+# NOTIFY_CHANNELS="apprise discord DSM generic gotify matrix ntfy-sh pushbullet pushover slack smtp telegram"
+#
+# DISABLE_DOCKCHECK_NOTIFICATION=false
+# DISABLE_NOTIFY_NOTIFICATION=false
+#
+# APPRISE_PAYLOAD='mailto://myemail:mypass@gmail.com
+#                 mastodons://{token}@{host}
+#                 pbul://o.gn5kj6nfhv736I7jC3cj3QLRiyhgl98b
+#                 tgram://{bot_token}/{chat_id}/'
+# APPRISE_URL="http://apprise.mydomain.tld:1234/notify/apprise"
+#
+# DISCORD_WEBHOOK_URL="https://discord.com/api/webhooks/<token string>"
+#
+# DSM_SENDMAILTO="me@mydomain.com"
+# DSM_SUBJECTTAG="Email Subject Prefix"
+#
+# GOTIFY_DOMAIN="https://gotify.domain.tld"
+# GOTIFY_TOKEN="token-value"
+#
+# MATRIX_ACCESS_TOKEN="token-value"
+# MATRIX_ROOM_ID="myroom"
+# MATRIX_SERVER_URL="https://matrix.yourdomain.tld"
+#
+# NOTIFY_TOPIC_NAME="YourUniqueTopicName"
+#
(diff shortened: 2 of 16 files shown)


mag37:
Very impressive and good work! 
I need to read it through properly and test it in a few different environments. 

I can hopefully test with Apprise, Telegram and some more.

@yoyoma2  @firmlyundecided might be open to test on DSM. 

This week is pretty hectic so I'm afraid this won't be worked on until Sunday at best. But I'll try to spin it through my Fedora, Alpine, Debian and Alma - environments asap. Should probably try to set up a OSX test too..

mag37:
This is my output testing the telegram - template:
```bash
./dockcheck.sh -x 10 -in
New version available! v0.6.5 ⇒ v0.6.4 
 Change Notes: ##Restructured update process - first pulls all images, then recreates all containers. Added -F option. 
entered

Sending dockcheck update notification - telegram

Sending notify_telegram.sh update notification - telegram
Sending notify_telegram.sh update notification - v2

Sending notify_v2.sh update notification - telegram
Sending notify_v2.sh update notification - v2
[##################################################] 20/20 

Containers on latest version:
!admiring_kepler - not checked, no compose file
beszel
dockge
filebrowser
flatnotes
forgejo
heimdall
mealie
metube
pingvin-share
snippet-box
tinyfilemanager
web1_nginx
web1_pg
web2_nginx
web3_apache
whoogle

Containers with updates available:
dozzle
homer
web2_mariadb

Sending telegram notification

No updates installed, exiting.
```

vorezal:
Thank you both for the feedback and thorough review!

To your points, yes my thinking here was that going forward we would recommend only making changes to the templates via environment variables in dockcheck.config. This would allow people to update configuration templates in place, possibly even automatically like the dockcheck.sh script itself, or pull the latest version of the repo while maintaining their configuration changes. If they wanted to completely modify things I think they would have to copy them to notify.sh to perform their own sort of personal override of the entire v2 setup. That might deserve its own callout in the templates. I only removed the line mentioning modification and didn't add anything actually saying _not_ to modify them.

I think I was intending for notify variables to cause a hard failure if they remained unset but added to the list of notification channels due to the -u flag, but I realize now this is being handled inconsistently at best. Plus that isn't the friendliest. A warning about channels being unset when -i is passed would be good, plus probably a check in each template to make sure the variables coming through are at least set. Let me look at that.

Agreed on putting the v2 readme section first if we're agreed this will be the new way!

vorezal:
I think I incorporated all the changes mentioned. I went with a warning and gracefully dropping notifications instead of a hard failure, but open to thoughts on that approach. Maybe it would be easier for folks to notice something was wrong and fix it if the script errors out completely, but this way if any channels are configured correctly it will run through and use them.

mag37:
I'll read it through properly and test it on my machines when I got some time to spare!

After a quick glance though,
I think this line should be changed in `notify_telegram.sh`:
```bash
# from:
  TelegramTopicID=12345678 ## Set to 0 if not using specific topic within chat
# to: 
  TelegramTopicID=${TELEGRAM_TOPIC_ID:="0"}
```
And also add that to the `default.config`:
```bash
# TELEGRAM_TOPIC_ID="0"
```

And also I'm wondering if there should be some more describing comments in the `default.config`? As currently you shouldn't even need to open/modify the `notify_v2.sh` file to get started, so people might not open it to read the comments there?
At least these two:
```bash
## Uncomment to not send notifications when dockcheck itself has updates.
# DISABLE_DOCKCHECK_NOTIFICATION=false
## Uncomment to not send notifications when notify scripts themselves have updates.
# DISABLE_NOTIFY_NOTIFICATION=false
```

And I dont know if it's just me being dense/weird but I also prefer having comments prefixed by 2 `#` and options by 1, to not uncomment raw text by mistake. Plus it a bit clearer separation if that makes sense.

mag37:
>Sorry about that! I completely forgot to address the Telegram miss.
No worries :) It's a bunch to keep track of! 

>I also never thought terribly carefully about the number of `#` and why that might be important, so good call.
Well I don't know about *important*, it's up to taste but I like it and have seen it handled similar in other config/option comments.

And no rush!

mag37:
Another one - just tested `notify_apprise.sh` (the API-variant) with v2 and it worked flawlessly - after fixing a faulty variable :)
At **line 27** change `$AppriseURL` to `$APPRISE_URL` 
Alternatively set it like you've done in other templates:  
```bash
if [[ -n "${APPRISE_URL:-}" ]]; then
AppriseURL="${APPRISE_URL}"
...
```

(starting to feel guilty I'm just posting everything as comments instead of just committing/PR'ing back to your fork)

vorezal:
Shoot thanks. No worries I'll get them addressed. They are my own mistakes anyway.

yoyoma2:
Looks good on DSM. Don't know about the other templates.

The only thing is the MessageBody in dockcheck_notification still doesn't end with \n (see yesterday's diff) but in notify_update_notification it does. Deleting it in notify_update_notification and adding it in trigger_DSM_notification after the MessageBody is also acceptable as long as MessageBody is consistent.

```
diff --git a/notify_templates/notify_DSM.sh b/notify_templates/notify_DSM.sh
index 17d697f..7c5d9f4 100644
--- a/notify_templates/notify_DSM.sh
+++ b/notify_templates/notify_DSM.sh
@@ -41,6 +41,7 @@ Content-Type: text/plain; charset=UTF-8; format=flowed
 Content-Transfer-Encoding: 7bit
 
 $MessageBody
+
  From $SenderName
 __EOF
 # This ensures DSM's container manager will also see the update
diff --git a/notify_templates/notify_v2.sh b/notify_templates/notify_v2.sh
index 8fb8cc0..b6d5b13 100644
--- a/notify_templates/notify_v2.sh
+++ b/notify_templates/notify_v2.sh
@@ -71,7 +71,7 @@ notify_update_notification() {
           if [[ "${!VersionVar}" != "$LatestNotifyRelease" ]] ; then
               MessageTitle="$FromHost - New version of notify_${notify_script}.sh available."
 
-              printf -v MessageBody "\nnotify_${notify_script}.sh update available:\n ${!VersionVar} -> $LatestNotifyRelease\n"
+              printf -v MessageBody "\nnotify_${notify_script}.sh update available:\n ${!VersionVar} -> $LatestNotifyRelease"
           fi
 
           printf "\n"
```

vorezal:
Yep I missed that as well. I'll get it fixed up. Thanks again!

mag37:
Just ran it trough my 4 test machines; arch, alpine, debian and fedora.
20 containers each and seems to work great (with the few tweaks listed above). **Great work!**

Though found another one :grin: a sneaky debug-echo `echo entered` at line 44 in `notify_v2.sh`.

vorezal:
Ah hah. Well perfection definitely eludes me still. Adding that to my list and will knock these out soon™. Overall though not too terrible for a decently large refactor. Thanks for the time and effort!

vorezal:
Alright the latest set of fixes is in.

- More comments in default.config with different `#` depth for comments and settings
- Renamed NOTIFY_TOPIC_NAME variable to NTFY_TOPIC_NAME for consistency
- Added TELEGRAM_TOPIC_ID
- Fixed AppriseURL variable
- Add an ending newline to all MessageBody statements for consistency
- Removed troubleshooting echo statement
- Prevented attempting to trigger notifications for notification templates if versions weren't different. Not sure how I missed this one.

yoyoma2:
Works fine on DSM. Even though I personally only use a single unmodified notify template, congrats @vorezal on a nice clean PR.

vorezal:
Thanks! It took some doing to get there, and a fair bit of help from you folks. I hope some find it useful.

mag37:
Well played @vorezal! And thank you for being a part of it @yoyoma2.

I'll do a another run of tests and then merge it!

vorezal:
Yep that's it. It should just be some harmless noise, but next round of fixes I would adjust line 358 of dockcheck.sh to eliminate the error output anyway.

```bash
 # Version check for notify templates
-[[ "$Notify" == true ]] && { exec_if_exists_or_fail notify_update_notification || printf "Could not source notify notification function.\n"; }
+[[ "$Notify" == true ]] && [[ ! -s "${ScriptWorkDir}/notify.sh" ]] && { exec_if_exists_or_fail notify_update_notification || printf "Could not source notify notification function.\n"; }
```
We could also set a variable to track when notify.sh was sourced vs notify_v2.sh, which might look a little cleaner in a few spots. It would pretty much just be a replacement for `[[ -s "${ScriptWorkDir}/notify.sh" ]]` though.

mag37:
I was thinking the same, just harmless info.
Good fix - I'll shove it in to main in a bit without bumping version for now.

And as everything else seems fine I'll plug a new release for v0.6.5 with info!

PR #183, opened 2025-05-24
add DaysOld to notify_ntfy-sh.sh

op4lat:
I've been using more often `-d` so I wanted notify_ntfy.sh.sh to tell me how many days I set with `-d`. It's less than 5 additional lines since dockcheck.sh already exports DaysOld so I don't even touch the main script.

<details>
  <summary>Spoiler</summary>
This is my second ever PR, so  I still not sure what I'm doing, Please let me know if the norm is to post an issue, the a PR
</details>

### DIFF
--- a/notify_templates/notify_ntfy-sh.sh
+++ b/notify_templates/notify_ntfy-sh.sh
@@ -1,9 +1,13 @@
 ### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
-NOTIFY_NTFYSH_VERSION="v0.2"
+NOTIFY_NTFYSH_VERSION="v0.3"
 #
 # Copy/rename this file to notify.sh to enable the notification snippet.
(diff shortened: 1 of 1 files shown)


vorezal:
Seems like a reasonable addition to me, though I'm sure mag37 will weigh in. Either way, there's an incoming change to how notifications work, so it will probably be easier to wait until that is in place. Once it is, I think this change would fit best in a slightly different spot and could apply to all notification templates.

mag37:
Yeah I agree - good idea and addition! 
But as @vorezal mentioned - a big change is on the way (hopefully merged tonight) and I think would fit elsewhere in a more generic execution. 

I wont decline the PR for now - but will keep it as a reminder to look at it and alternative implementation of the idea asap.

Thank you for contributing!

op4lat:
Hello both. Okay, sounds good!

PR #188, opened 2025-05-27
Notify_v2.sh bug fixes

vorezal:
* Clarify notify_v2.sh usage in README.md

* Fix JSON newline handling in Discord and Telegram channels

* Additional error messages when notification templates fail to be sourced

* Additional variable for self-hosted ntfy.sh domain

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,27 @@ Run it scheduled with `-ni` to only get notified when there's updates available!
 
 V2 installation and configuration (tag v0.6.5 or later):
 Remove or rename `notify.sh` if previously configured using the legacy method.
+Make certain your project directory is laid out as below. You only need the notify_v2.sh file and any notification templates you wish to enable, but there is no harm in having all of them present.
+```
+ .
+├── notify_templates/
+│   ├── notify_DSM.sh
+│   ├── notify_apprise.sh
+│   ├── notify_discord.sh
+│   ├── notify_generic.sh
+│   ├── notify_gotify.sh
+│   ├── notify_matrix.sh
+│   ├── notify_ntfy-sh.sh
+│   ├── notify_pushbullet.sh
+│   ├── notify_pushover.sh
+│   ├── notify_slack.sh
+│   ├── notify_smtp.sh
+│   ├── notify_telegram.sh
+│   └── notify_v2.sh
+├── dockcheck.config
+├── dockcheck.sh
+└── urls.list         # optional
+```
 Uncomment and set the NOTIFY_CHANNELS environment variable in `dockcheck.config` to a space separated string of your desired notification channels to enable.
 Uncomment and set the environment variables related to the enabled notification channels.
 It is recommended not to make changes directly to the `notify_X.sh` template files and to use only environment variables defined in `dockcheck.config` using this method.
--- a/default.config
+++ b/default.config
@@ -54,6 +54,8 @@
 # MATRIX_ROOM_ID="myroom"
 # MATRIX_SERVER_URL="https://matrix.yourdomain.tld"
 #
+# ntfy.sh or your custom domain with no trailing /
+# NTFY_DOMAIN="ntfy.sh"
 # NTFY_TOPIC_NAME="YourUniqueTopicName"
 #
 # PUSHBULLET_URL="https://api.pushbullet.com/v2/pushes"
--- a/notify_templates/notify_discord.sh
+++ b/notify_templates/notify_discord.sh
@@ -1,5 +1,5 @@
 ### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
-NOTIFY_DISCORD_VERSION="v0.2"
+NOTIFY_DISCORD_VERSION="v0.3"
 #
 # Required receiving services must already be set up.
 # Do not modify this file directly. Set DISCORD_WEBHOOK_URL in your dockcheck.config file.
@@ -13,6 +13,10 @@ fi
 trigger_discord_notification() {
   DiscordWebhookUrl="${DISCORD_WEBHOOK_URL}" # e.g. DISCORD_WEBHOOK_URL=https://discord.com/api/webhooks/<token string>
 
-  MsgBody="{\"username\":\"$FromHost\",\"content\":\"$MessageBody\"}"
-  curl -sS -o /dev/null --fail -X POST -H "Content-Type: application/json" -d "$MsgBody" "$DiscordWebhookUrl"
+  JsonData=$( jq -n \
+              --arg username "$FromHost" \
+              --arg body "$MessageBody" \
+              '{"username": $username, "content": $body}' )
+
+  curl -sS -o /dev/null --fail -X POST -H "Content-Type: application/json" -d "$JsonData" "$DiscordWebhookUrl"
 }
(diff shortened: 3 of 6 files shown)


vorezal:
Trying to make up for my mistakes :sweat_smile: Let me know what you think of the documentation changes in particular, and a Telegram test would be great if anyone is able. I was able to spin up my own Discord server and my testing there looked good, but a second look never hurts either.

vorezal:
Additionally, what would thoughts be on this documentation and/or a combination of supporting pulling notify templates from the project root directory for people that still want to override and then falling back to the subdirectory?

mag37:
> Hmm the TelegramToken variable should have been set. It's pulling it to the local variable on line 20: `TelegramToken="${TELEGRAM_TOKEN}" # e.g. TELEGRAM_TOKEN=token-value` Did it fail to execute regardless? It is probably unnecessary when we can just use the variable directly from the config, but it might make it (very slightly) easier for people that do direct modifications.

You're right - must've been something I messed up! I just did another test run with a clean git clone and it worked fine, so I've must have messed something up in my testing. Sorry about the confusion - I also think it's good to keep the current setup to allow for simple modification.

> And yes, I mean we could set up the notify_v2 function to check the project root for any template files first, then fall back to the notify_templates folder. I like that as well, so let me write that up. It will require some additional tweaking to the documentation (and .gitignore) though.

Great! I think that's a good direction, I'll hold on merging this until we've ironed out some more things then.

> My mistake on the config file comment consistency. Will try to keep that consistent for sure!

No worries!


@xconverge 
> Just wanted to chime in and say the refactor of notifications looks great, so no need to "make up for your mistakes" !!! Thanks all

I totally agree (and thank you for your support) -  @vorezal did some great work here! Just minor tweaks and fixups to get it all in line, but great work either way.

vorezal:
Yep I should be able to get that added pretty easily with the rest of this. Will do!

vorezal:
Totally fair points. I'd prefer to be picky and get it right. The comment in the template files was just getting pretty long, but in this case clarity is probably best.

mag37:
Yeah, 2 lines looks good and clear! 
Nice one.

I'll do another version bump and release when this is done for merging. Great work again.

vorezal:
OK hopefully that's the last of the fixes (for now). I did use slightly different language for the "days old" message feature than used in https://github.com/mag37/dockcheck/pull/183, but just to match the text used in the -d flag's help text more closely. Hopefully that still works for @op4lat.

vorezal:
It looks like in some other notification channels (including all of the ones I have available for testing unfortunately) the extra newline is just squashed. I don't see any change in the output on those, so I would default to removing the extra newline for channels that display it.

On a somewhat related note, I just had a notification to Slack hang for several minutes without failing. I know we added a retry for script updates, so maybe it would make sense to add retries and a lower max time per request for most/all curl calls in the scripts? If there were a standard set of arguments we could set a variable in one spot and pass it to all of the curl requests.

mag37:
>If there were a standard set of arguments we could set a variable in one spot and pass it to all of the curl requests.

That sounds reasonable! I just guesstimated some values when I did the curl/wget changes in the main script
This is currently used:
```bash
curl --retry 3 --retry-delay 1 --retry-max-time 10 
```
Which boils down to:
- `--retry 3` = tries 3 times
- `--retry-delay` = waits 1 second between retries (if omitted curl defaults to exponential -> 1s,2s,4s)
- `--retry-max-time 10` = total max time for retrying

Though maybe it should be more sensible running something like:
```bash
curl --retry 3 --retry-delay 1 --connect-timeout 5
```
Which would mean 5 retries with 1 second in between each and a timeout of 5 for each try, max ~15s.

But yes - having it as one or more global variables configurable by the user is a good idea!

op4lat:
> OK hopefully that's the last of the fixes (for now). I did use slightly different language for the "days old" message feature than used in #183, but just to match the text used in the -d flag's help text more closely. Hopefully that still works for @op4lat.

Thanks for the acknowledge, lol. I  am still checking out the new feature and looks cool. I have upgraded all my containers so I don't have any 'days old' to debug, yet. 

I did notice that notify_ntfy.sh is hardwired to ntfy.sh so I'll submit a PR to have a custom NTFY_DOMAIN for us that host our own ntfy.

vorezal:
I removed the extra newline, though I did it from the UpdToString variable. Everywhere else sets a newline explicitly at the end of MessageBody, so I figure it makes sense to maintain that for consistency.  Let me know if it would be preferred to remove it from the MessageBody variable instead though.

vorezal:
All good on my side I think. I have some other things to contribute, but the scope here seems reasonable as a bugfix PR. I'll add anything beyond this as new features. If I get to any of those soon enough I may look into standardizing some curl arguments and including that at the same time. Assuming you or someone else doesn't get to it first anyway.

mag37:
Another one (or two) just surfaced in #193 

- first the `jq` call should be made with the `jqbin` variable.
- second - `hostname` command is not as universal as `cat /etc/hostname` so that's a reasonable change to add.

mag37:
> The newline suffix stripping works perfectly on DSM. 👍 You added curly braces on UpdToString when printing MessageBody but LatestNotifyRelease has none in notify_update_notification when similarly printing MessageBody. Just curious... On DSM curly baces or not are both working on that line. I'm all for improving code readability.

Glad to hear it's working! Then that bit is done :white_check_mark: 

The curly braces (if I guess) is just due to **me** being inconsistent in the original code - in my own opinion curly braces is preferred and makes for safer code when the variable is within other strings like `${One}TwoThree` and is required when expanding array elements or doing parameter expansion.
I've tried to think that a stand alone variable like `This is the $element` might be safe but one in eg. a path or url like `/path/to/variable/${directory}/one` is sensible to encase in curly braces.

op4lat:
Merge away!

yoyoma2:
I even tried `cat /etc/hostname` on DSM and it's fine. Overkill maybe but DSM keeps surprising me by how different it is from normal distributions. Merge away!

mag37:
I just realized that this newline -fix might cause issues with the legacy `notify.sh` CLI output.

It's already merged to main and it's not breaking (just ugly) - I'll troubleshoot it further tomorrow.
Otherwise it seems to be working great, well done @vorezal ! And thank you for staying on and continue working.

vorezal:
Well shoot. We only made the change to notify_v2.sh though, so I thought any existing copies of notify.sh would be unaffected. What seems to be the trouble exactly? Is it when people might copy notify_v2.sh to notify.sh themselves?

mag37:
> Well shoot. We only made the change to notify_v2.sh though, so I thought any existing copies of notify.sh would be unaffected. What seems to be the trouble exactly? Is it when people might copy notify_v2.sh to notify.sh themselves?

You're right, shouldn't affect legacy.
Was just among the last few test runs after patching I saw it on one machine, but it was probably some mismatch as I just shuffled around old configs and `notify.sh` to try to find edge cases.

I'll do another check - was too tired to trust any troubleshooting!

mag37:
I can't reproduce it! So I guess I just had some odd template laying around in my test-files - but the output was something like:

```bash
...
Containers with updates available:
apprise-api
caddy
dozzle
homer
metube
apprise-api\ncaddy\ndozzle\nhomer\nmetube\n
...
```
So it seemed very related - though now I've tried a bunch of different config combinations:
- legacy notify.sh + no .config
- legacy notify.sh + new settings in .config
- v2 + settings in .config and template in subdir
- v2 + settings in .config and template in rootdir
- v2 + no settings -> correctly gives me error message.

So it seems all good!

yoyoma2:
I'm running 0.6.6 with notify v2 enabled and it works just as well as notify.sh used to. Congrats!

In the refactor PR @vorezal said "This would allow people to update ... possibly even automatically like the dockcheck.sh script itself".

Now that notification customization is done in the config file (beautiful!) it now makes sense to allow automatically updating the scripts in notify_templates just like the dockcheck.sh script itself.

PR #198, opened 2025-06-01
delete show-error on curl

op4lat:
Not sure if this was intentional or not, but while I was trouble shooting I noticed that the curl command in notify_ntfy.sh uses `curl -sS --show-error` and `-S` is short for `--show-error` so figure that either should be deleted. My PR deletes `--show-error` from the other templates:

- notify_ntfy.sh 
- notify_pushbullet.sh
- notify_pushover.sh
- notify_slack.sh


mag37:
Ah you're correct! Good find.

I might hold back on this PR some as I intend to restructure the curl/wget commands to have their timeout settings configurable in the config.

PR #200, opened 2025-06-03
Snooze feature, curl, and consolidation

vorezal:
* Added snooze feature to delay notifications

* Added configurable default curl arguments

* Consolidated and standardized notify template update notifications

* Added curl error handling

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,6 @@ ___
       - Added support for notification management via environment variables.
       - Moved notification secrets to **dockcheck.config**.
     - Added retries to wget/curl to not get empty responses when github is slow.
-- **v0.6.4**: Restructured the update process - first pulls all updates, then recreates all containers.
-    - Added logic to skip update check on non-compose containers (unless `-r` option).
-    - Added option `-F` to revert to `compose up -d <ContainerName>` targeting specific container and not the stack.
-        - Also added corresponding label and config-option.
-    - Added markdown formatting to `notify_ntfy.sh` template.
-- **v0.6.3**: Some fixes and changes:
-    - Stops when a container recreation (compose up -d) fails, also `up`s the whole stack now.
-    - `-M`, Markdown format url-releasenotes in notification (requires template rework, look at gotify!)
-    - Added [addons/DSM/README.md](./addons/DSM/README.md) for more info Synology DSM info.
-    - Permission checks - graceful exit if no docker permissions + checking if root for pkg-manager.
 ___
 
 
@@ -142,7 +132,7 @@ If `notify.sh` is present and configured, it will be used. Otherwise, `notify_v2
 Will send a list of containers with updates available and a notification when `dockcheck.sh` itself has an update.
 Run it scheduled with `-ni` to only get notified when there's updates available!
 
-Installation and configuration:
+#### Installation and configuration:
 Make certain your project directory is laid out as below. You only need the notify_v2.sh file and any notification templates you wish to enable, but there is no harm in having all of them present.
 ```
  .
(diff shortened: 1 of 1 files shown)


mag37:
Sounds like you've done a thorough job indeed, nice with cleaning up the other flags too.

About the `--retry-max-time` and `--max-time` flags:  

I think the `--retry-max-time` is redundant, especially if we use the `--connect-timeout` instead to handle the timout on actually establishing the connection, curl docs:
>The connection timeout only limits the time curl is allowed to spend up until the moment it connects, so once the TCP connection has been established it can take a longer time.

And `--max-time 5` will cut the connection after 5 seconds no matter what's going on - so a slow connection wont have time to complete the session (including DNS, TLS handshake, downloading and processing).   

I think `--max-time` should be replaced with `-connect-timeout` instead - to break if no successful connection is made within the timeframe but allow for longer running downloads to account for slower connections to be able to complete.

>I added the curl parameters to a variable in the dockcheck.config file, so all good there.

Ah yes sorry - missed that in my previous glance. :) 
Though in **dockcheck.sh** where you set the default values, is there a reason you chose `-` over `=` in the parameter expansion `${CurlArgs:-...` ?  I've used `=` but maybe I'm wrong in how it works.

My understanding is that `=` also sets the expanded variable to that value, which doesn't matter as that's what we're doing anyway by setting `Var123=${Var123:="default"}` but it also sets to the default value when the variable is set to null, like `Var123=""` will be replaced by the default value if used with `=` but not with `-`.

mag37:
True, having the arguments "open" to fill with whatever you want could be nice for those who want to tweak it specifically - but it's also a balance in the possibility of breakage and hard troubleshooting :grin:  I'm not set on either - it was just a suggestion/alternative. 

And that's a great little table showing the parameter expansions! And with more careful reading it seems like `:-` and `:=` should work the same except for the assignment. Which indeed is redundant in this case. So maybe all variable initialization under `# Initialise variables` should be replaced to `:-` to be

vorezal:
I've pushed some adjustments with most of what was discussed above for review.

Also, thinking about it further (plus realizing I completely missed the part where you stated only curl is supported in a number of spots, which is obvious now but I didn't even think about it before), it does make a fair amount of sense to deprecate wget. If making that a dependency though it seems it would be pretty difficult to support a binary installation since those are only hosted by third parties in separate spots for each Linux distribution, and building from source automatically would probably be equally problematic if not more so. Perhaps add some logic that prompts to install, only support package manager installation, and if it is not successful we disable any curl based functionality instead of failing entirely? That should just disable script update checks and notifications.

mag37:
I've not yet had time to properly read through or test it, but looks like you covered the discussion in the latest commit.

A question - do you need the `cat` at 105 in `notify_v2.sh`  ?
```bash
  done <<< "$(cat ${SnoozeFile} | grep ${switch} '\.sh ')"
```
Cant you just `  done <<< "$(grep ${switch} '\.sh ' ${SnoozeFile})"` 

And about curl.

>If making that a dependency though it seems it would be pretty difficult to support a binary installation ...

Agreed - I don't mean as a dependency that `dockcheck.sh` will help you install - but a requirement to run dockcheck at all.

> ... but I've definitely come across distros, mostly Docker images, that include wget and not curl out of the box.

So I'm considering we either build it so that all curl uses has a wget alternative or drop wget completely and as you mention - give a message about skipped functionality instead of failing completely.

mag37:
> Is there something I can set in snooze seconds to avoid duplicate notifications forever? Can I target different containers differently? Some containers I want to be reminded more some less some very little based on importance and use case.

If a per container specific setting is wanted - maybe that's more sensible to control with compose labels rather than configuring in a config file? Though that'd mean that we'd need to extract the labels for each container before doing any notification logic.
Sounds a little over the top :)

yoyoma2:
> I don't have anything written to delay or prevent the initial notification. Just making sure that was clear.

Yes that was clear, only talking about repeat notifications. To save people some snooze seconds calculations you could use some predefined definitions [like cron does](https://en.wikipedia.org/wiki/Cron#Nonstandard_predefined_scheduling_definitions) for common durations and the -1 infinite snooze.

mag37:
>So, all that to say that adding container specific snooze settings wouldn't really do much without some changes. Other containers triggering notifications, either due to having a new update or their snooze windows expiring, would include the containers with longer snooze windows anyway.

I must say that I think having snooze-settings pre container opens up a whole jar of puzzles to solve and might be very confusing to configure - and would require a lot of more code and processing. And I don't think it's the right direction.

One option could be to have a option flag to override/skip the snooze settings?  
That way you could have your scheduled run with a snooze-setting from the config. And if you have a specific container you'd like to notify straight away no matter what, you could have a different scheduled run with eg `dockcheck.sh -nix 10 -Z very_important_container` (where Z is skip-snooze) and it would trigger an instant notification as usual.

Just an idea - but I don't see the use case for this detailed configuration per container in general to be worth the potential issues and large rewrite.

mag37:
And to continue on that - currently it's only possible to pass one name as the filter. Even though it wildcards, so passing nginx like `dockcheck.sh nginx` would check for **nginx-web-main** and **reverse-nginx** etc.

A modification I've considered before is to rewrite it to allow multiple (but exact) matches instead.  
So the rewrite would be:
```bash
# from:
  docker ps $Stopped --filter "name=$SearchName" --format '{{.Names}}'
# to:
  docker ps $Stopped --filter "name=^(${SearchName//,/|})$" --format '{{.Names}}'
```
Which would then make it possible to run it like `dockcheck.sh -nix 10 cont1,another2,third`.
To filter exact which ones you'd like to update.

This is a bit out of this scope - but a change I'll probably implement either way.

vorezal:
I still have a few things to tweak I think. I've just been busy and haven't gotten to it yet. Hopefully soon!

mag37:
> I still have a few things to tweak I think. I've just been busy and haven't gotten to it yet. Hopefully soon!

No rush at all! Was just considering if something was waiting on me ^^ take your time.

mag37:
True!
Either way, if supporting the name also clarifying the readme some (maybe make it shorter but clearer and only briefly mention legacy?).

yoyoma2:
What if notify_v2.sh later becomes notify_v3.sh, notify_v4.sh, notify_v5.sh... will many hardcoded checks be required? A documentation change specifying clearly to use notify.sh and this issue disappears forever.

mag37:
> What if notify_v2.sh later becomes notify_v3.sh, notify_v4.sh, notify_v5.sh... will many hardcoded checks be required? A documentation change specifying clearly to use notify.sh and this issue disappears forever.

This is a good point and I agree! 
That makes it possible to just rename the v2 to something more generic like "wrapper" or so and completely phase out legacy over time too.

yoyoma2:
> That makes it possible to just rename the v2 to something more generic like "wrapper" or so and completely phase out legacy over time too.

Even better.

vorezal:
Good points! Yes, perhaps it will be better long term to just clarify in the documentation. I'll go with that as soon as I can put a little bit more time in to finish this all up.

vorezal:
It's been a bit and I can't recall if we were going to implement any changes other than the unnecessary cat and readme clarifications. If I missed anything you can think of let me know, and please critique my readme changes. Hopefully things are at least a little bit simplified and more clear, but I'm reading these from the perspective of someone who knows what they do, which makes it difficult.

As for the feature itself, I think I would vote for leaving it as is for the moment (with an admittedly pretty specific use case) and I'll look into extending it in the future. Thanks!

mag37:
Read through it all briefly and looks good - looks like you've covered all we discussed as current changes.

I'll run through some tests later tonight or tomorrow and will read the readme-bits again if it seems clear enough.
If it all seems fine I'll merge it to main asap - might add some minor tweaks post-merge and push it as a release.

PR #207, opened 2025-06-25
Add DisplaySourcedFiles variable

op4lat:
This PR shows what config files are being sourced. DisplaySourcedFiles is added to toggle this feature. Read "feature" as me being picky so I won't be offended if it isn't approved.

```
$ ./dockcheck.sh -x30 -ni
 * /tmp/1/dc2/dockcheck/dockcheck.config is being used
 * /tmp/1/dc2/dockcheck/notify.sh is being used
New version available! v0.6.8 ⇒ v0.6.7 
 Change Notes: snooze feature (see readme), curl arguments, cleanup. 

Sending ntfy.sh dockcheck notification
```

### DIFF
--- a/default.config
+++ b/default.config
@@ -26,6 +26,7 @@
 #CurlRetryDelay=1    # Time between curl retries
 #CurlRetryCount=3    # Max number of curl retries
 #CurlConnectTimeout=5    # Time to wait for curl to establish a connection before failing
+#DisplaySourcedFiles=false    # Display what files are being sourced/used
 
 ### Notify settings
 ## All commented values are examples only. Modify as needed.
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -17,7 +17,7 @@ ScriptWorkDir="$(dirname "$ScriptPath")"
 source_if_exists_or_fail() {
   if [[ -s "$1" ]]; then
     source "$1"
-    [[ "${DisplaySourcedFiles:-false}" == true ]] && echo " * ${$1} is being used"
+    [[ "${DisplaySourcedFiles:-false}" == true ]] && echo " * sourced config: ${1}"
     return 0
   else
     return 1
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
-VERSION="v0.6.8"
-# ChangeNotes: DisplaySourcedfiles variable displays what configuration files are being used 
+VERSION="v0.6.7"
+# ChangeNotes: snooze feature (see readme), curl arguments, cleanup.n files are being used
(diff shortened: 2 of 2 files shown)


op4lat:
```
# Source helper functions
source_if_exists() {
  srcFile="$1"
  if [[ -s "$srcFile" ]]; then
    source "$srcFile"
    DisplaySourcedFiles=${DisplaySourcedFiles:-false}
    [[ "$DisplaySourcedFiles" == true ]] && echo " * ${srcFile} is being used"
  fi
}

source_if_exists_or_fail() {
  srcFile="$1"
  if [[ -s "$srcFile" ]]; then
    source "$srcFile"
    DisplaySourcedFiles=${DisplaySourcedFiles:-false}
    [[ "$DisplaySourcedFiles" == true ]] && echo " * ${srcFile} is being used"
  else
    return 1
  fi
}
```

Glad to hear that the idea is welcome! And I let you guys decide about whether using `srcFile` or not, but I am using it because I remember a time where using `"$1"` many times, the variable lose its value. Was it Bash in iOS? or was it busybox? cannot remember, but either way is fine.

Now,  I think `DisplaySourcedFiles=${DisplaySourcedFiles:-false}` should stay, because no variable is declared yet. Remember, by default `DisplaySourcedFiles` is commented out in the config file, so it isn't never initialized and I'm making sure that at least it is false by default. Trust me, I wanted to give it colors (sorry, I live in the other side of the pound) but I learned that nothing has been declared yet so didn't want the code to look uglier with ANSI color codes.

About using a flag, also your call, but I feel that another feature could use whatever letters are left.

mag37:
> My thinking on the config file output would be that it would tell people if we were sourcing the file from their home directory or from the dockcheck project directory. If that isn't really helpful or a concern, then yeah no need to output that one at all.

Yeah, that's true though still pretty easy to troubleshoot.
I'm not saying it's not worth considering, just that it might clutter more than it would help currently.

vorezal:
I just ran into this in another thread, but we may need to explicitly return 0 as the final line in the cases we successfully source the files (after the echo). If `[[ "$DisplaySourcedFiles" == true ]]` evaluates to false I think the function could begin to act like the sourcing actually failed rather than we just didn't want the output.

mag37:
Good work.
I did some testing and seems to be working fine, I'm not seeing the `source_if_exists()` function called anywhere else and seem to be working fine with the removal you've done.

@vorezal might have some input - otherwise I'll merge it as is.

vorezal:
Yes I must have thought it would be useful to avoid an error response code in some cases, but the use case doesn't really seem to exist anymore. I think there were some spots in nofity_v2.sh, but they probably got refactored out. All good to remove as far as I can tell.

Issue #208, opened 2025-06-25
Unbound variable -> "notify_templates/notify_v2.sh: line 235: Updates: unbound variable"


mag37:
> Just updated to v0.6.7 and get an error running `sudo dockcheck.sh -n -i -r`
> 
> notify_templates/notify_v2.sh: line 235: Updates: unbound variable
> 
> The following line is the culprit:
> `if [[ "${#Updates[@]}" -gt 0 ]]; then`
> 
> The releasenotes() function does an `unset Updates`
> 
> Does this [stackoverflow answer](https://stackoverflow.com/a/61551944) help getting a portable check for an empty array? 

 _Originally posted by @yoyoma2 in [#150](https://github.com/mag37/dockcheck/issues/150#issuecomment-3004920608)_

Thank you @yoyoma2 I've must have missed this in my testing somehow - sorry.  

I'll try to debug it - the "unset Updates" in the releasenotes function does not seem to be related though from my initial glance.

@vorezal any idea if you've got time to look at it?

vorezal:
I may be able to take a look tonight hopefully.

mag37:
I'll note some things while troubleshooting - I'll fix them later if there's no objections.

First a variable mismatch
- `DISABLE_NOTIFY_NOTIFICATION` in the default.config
- `DISABLE_NOTIFY_UPDATE_NOTIFICATION` in the notify_v2.sh
I'd suggest we settle for the shorter.

Then there's a few places in the **notify_v2.sh** where a check if variables are true is done with a quoted `"true"` instead of an unquoted `true` which I think should be consistent with the rest of the script (which uses unquoted, to not mistakenly handle as a string). Current example:
```
if [[ ! "${DISABLE_NOTIFY_UPDATE_NOTIFICATION:-}" == "true" ]]; then
```

mag37:
> Yes I should be able to get a quick fix in for this. I probably shouldn't even be using the "Updates" variable at all to avoid confusion with the global variable the main script uses. I'll use something else for this function and make sure to initialize it correctly.

Sounds reasonable. With some quick testing of just renaming variables and presetting the variable like `UpdatesTest=()` within the `notify_update_notification()` function I got rid of the unbound variable - but it's throwing a **Could not source notification function.**.

I'll continue troubleshooting as soon as I can!

vorezal:
Whoops! We can go with the shorter variable. I just thought "notify notification" read strangely.

[...]

PR #209, opened 2025-06-26
Fix unbound variable, potential collision, and config variable.

vorezal:
- Fix potentially unbound "Updates" variable and prevent possible global collision
- Fix mismatched DISABLE_NOTIFY_NOTIFICATION variable between config file and notify_v2.sh script

### DIFF
--- a/notify_templates/notify_v2.sh
+++ b/notify_templates/notify_v2.sh
@@ -139,6 +139,8 @@ send_notification() {
   fi
 
   [[ -n "${snooze}" ]] && cleanup_snooze "${Updates[@]}"
+
+  return 0
 }
 
 ### Set DISABLE_DOCKCHECK_NOTIFICATION=false in dockcheck.config
@@ -187,6 +189,8 @@ dockcheck_notification() {
       fi
     fi
   fi
+
+  return 0
 }
 
 ### Set DISABLE_NOTIFY_NOTIFICATION=false in dockcheck.config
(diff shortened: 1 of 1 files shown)


vorezal:
@mag37 addresses issue #208

vorezal:
Ah the touch I can put behind the flag to enable snooze if it's not already. I wasn't able to get that source error when I was testing. What is that Notif_notif and Update_notif output? Also, it seems like both notifications work based on the output and I thought everything was sourced earlier in the script than that.

mag37:
> Ah the touch I can put behind the flag to enable snooze if it's not already. I wasn't able to get that source error when I was testing. What is that Notif_notif and Update_notif output? Also, it seems like both notifications work based on the output and I thought everything was sourced earlier in the script than that.

Ah sorry I forgot to say that I added a prefix before the three different error messages to which one (or all) that was throwing the error. Just while I was debugging, shoul've cleaned that from the output.

```bash
# line 359
       [[ "$Notify" == true ]] && { exec_if_exists_or_fail dockcheck_notification "$VERSION" "$LatestRelease" "$LatestChanges" || printf "Could not source notification function.\n"; }

# line 367:
 [[ "$Notify" == true ]] && [[ ! -s "${ScriptWorkDir}/notify.sh" ]] && { exec_if_exists_or_fail notify_update_notification || printf "Notif_notif: Could not source notify notification function.\n"; }

# line 504
   [[ "$Notify" == true ]] && { exec_if_exists_or_fail send_notification "${GotUpdates[@]}" || printf "\nCould not source notification function.\n"; }
```

These are the three places that's throwing the message, even though both the notifications actually works.

vorezal:
Good find! That must be it. It's throwing an error, not actually failing to source. Thanks, I'll figure out a fix.

PR #213, opened 2025-07-11
first version of notification to Home Assistant

TermeHansen:


TermeHansen:
@mag37 something like this? Feel free to update/change :smile_cat:

PR #219, opened 2025-07-31
update SMTP template, added suport for sendmail

xmirakulix:
I use the sendmail binary on my systems for mail delivery, added support in the notify_smtp.sh for this as well.

### DIFF
--- a/notify_templates/notify_DSM.sh
+++ b/notify_templates/notify_DSM.sh
@@ -1,5 +1,5 @@
 ### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
-NOTIFY_DSM_VERSION="v0.31"
+NOTIFY_DSM_VERSION="v0.4"
 # INFO: ssmtp is deprecated - consider to use msmtp instead.
 #
 # mSMTP/sSMTP has to be installed and configured manually.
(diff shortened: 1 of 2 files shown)


yoyoma2:
Nice and simple and useful.

Wondering if a ".1" minor version should be appended to the NOTIFY_SMTP_VERSION?

Wondering if the same changes should be added to the DSM notification template which is almost identical?

Just an observer here so wait for official comments from @mag37.

mag37:
Wonderful! Thank you.

Yes thank you @yoyoma2 - A version bump to trigger update logics would be nice. 

And yes, if sendmail is a possibility on the DSM I think it's a good idea to add it to that template too.

xmirakulix:
both done 👍🏻

mag37:
Ah thats good for DSM then.

Yeah, I think `0.4` would be correct - thats how the other templates have incremented.

I will do some edits to the readme later and can add this. Good point.

Should probably edit this line too: 
```sh
## From:
         echo "No msmtp or ssmtp binary found in PATH: $PATH" ; exit 1
## To:
         echo "No msmtp, ssmtp or sendmail binary found in PATH: $PATH" ; exit 1
```

mag37:
> PS: I'm not entirely sure the individual versioning of the notify scripts makes much sense at all...

It's set up that way so if your template changes you can get notified that it got an update, decoupled from the main script. 
Goes both ways - to not force a new version/release for every minor tweak to a template nor making you check your notification with every new release :) 


Thank you a lot for the contribution and the edits! 
Merging to main.

yoyoma2:
Got a notification for the DSM template today so I figured this got merged. 👍 Love this tool!

My software development experience is extensive but my docker knowledge is not. Being only on the sidelines for a docker tool is appropriate.

PR #222, opened 2025-08-14
Add file notification channel

vorezal:
Adds a new "file" notification channel that writes out to a file in the main script project directory. This is intended to slightly more officially support the API implementation discussed here: https://github.com/mag37/dockcheck/discussions/146, although I'm pretty sure the Python API implementation itself is outside the scope of anything that would be included in this repository.

- Writes a comma separated list of container names (and release notes URLs, if enabled) with updates available to updates_available.txt in the main script directory
- Bypasses snooze to make sure the file and API always have the most up-to-date information available
- Bypass the file channel for other notification types (there may be a better, more flexible way of doing this)
- Adds a notify_v2+ exclusive call when there are no updates available to keep the snooze and updates files current
- .gitignore update for updates_available.txt file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 regctl
 # ignore snooze file
 snooze.list
+# ignore updates file
+updates_available.txt
--- a/default.config
+++ b/default.config
@@ -32,7 +32,7 @@
 ## All commented values are examples only. Modify as needed.
 ##
 ## Uncomment the line below and specify the notification channels you wish to enable in a space separated string
-# NOTIFY_CHANNELS="apprise discord DSM generic HA gotify matrix ntfy pushbullet pushover slack smtp telegram"
+# NOTIFY_CHANNELS="apprise discord DSM file generic HA gotify matrix ntfy pushbullet pushover slack smtp telegram"
 #
 ## Uncomment the line below and specify the number of seconds to delay notifications to enable snooze
 # SNOOZE_SECONDS=86400
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
-VERSION="v0.7.0"
-# ChangeNotes: Snooze bugfix, added auth support to ntfy.sh and sendmail support to SMTP
+VERSION="v0.7.1"
+# ChangeNotes: Call send_notification even when no updates are available to write to file (if enabled) and clean up snooze
 Github="https://github.com/mag37/dockcheck"
 RawUrl="https://raw.githubusercontent.com/mag37/dockcheck/main/dockcheck.sh"
 
@@ -507,6 +507,8 @@ if [[ -n ${GotUpdates[*]:-} ]]; then
   if [[ -s "$ScriptWorkDir/urls.list" ]] && [[ "$PrintReleaseURL" == true ]]; then releasenotes; else Updates=("${GotUpdates[@]}"); fi
   [[ "$AutoMode" == false ]] && list_options || printf "%s\n" "${Updates[@]}"
   [[ "$Notify" == true ]] && { exec_if_exists_or_fail send_notification "${GotUpdates[@]}" || printf "\nCould not source notification function.\n"; }
+else
+  [[ "$Notify" == true ]] && [[ ! -s "${ScriptWorkDir}/notify.sh" ]] && { exec_if_exists_or_fail send_notification "${GotUpdates[@]}" || printf "\nCould not source notification function.\n"; }
 fi
 
 # Optionally get updates if there's any
--- a/notify_templates/notify_file.sh
+++ b/notify_templates/notify_file.sh
@@ -1,0 +1,12 @@
+### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
+NOTIFY_FILE_VERSION="v0.1"
+#
+# Leave (or place) this file in the "notify_templates" subdirectory within the same directory as the main dockcheck.sh script.
+# If you instead wish make your own modifications, make a copy in the same directory as the main dockcheck.sh script.
+
+trigger_file_notification() {
+  NotifyFile="${ScriptWorkDir}/updates_available.txt"
+
+  echo "${MessageBody}" > ${NotifyFile}
+
+}
--- a/notify_templates/notify_v2.sh
+++ b/notify_templates/notify_v2.sh
@@ -114,6 +114,23 @@ send_notification() {
     UpdNotifyCount="${#Updates[@]}"
   fi
 
+  if [[ "${enabled_notify_channels[@]}" == *"file"* ]]; then
+    UpdToString=$( printf '%s, ' "${Updates[@]}" )
+    UpdToString="${UpdToString%, }"
+
+    if [[ -z "${UpdToString}" ]]; then
+      UpdToString="None"
+    fi
+
+    printf "\nSending file notification\n"
+    printf -v MessageBody "${UpdToString}"
+
+    exec_if_exists_or_fail trigger_file_notification || \
+    printf "Attempted to send notification to channel file, but the function was not found. Make sure notify_file.sh is available in the ${ScriptWorkDir} directory or notify_templates subdirectory.\n"
+
+    remove_channel file
+  fi
+
   NotifyError=false
 
   if [[ "${UpdNotifyCount}" -gt 0 ]]; then
@@ -238,9 +255,11 @@ notify_update_notification() {
         printf -v MessageBody "Notify templates on $FromHost with updates available:\n${UpdToString}\n"
 
         for channel in "${enabled_notify_channels[@]}"; do
-          printf "Sending notify template update notification - ${channel}\n"
-          exec_if_exists_or_fail trigger_${channel}_notification || \
-          printf "Attempted to send notification to channel ${channel}, but the function was not found. Make sure notify_${channel}.sh is available in the ${ScriptWorkDir} directory or notify_templates subdirectory.\n"
+          if [[ ! "${channel}" == "file" ]]; then
+            printf "Sending notify template update notification - ${channel}\n"
+            exec_if_exists_or_fail trigger_${channel}_notification || \
+            printf "Attempted to send notification to channel ${channel}, but the function was not found. Make sure notify_${channel}.sh is available in the ${ScriptWorkDir} directory or notify_templates subdirectory.\n"
+          fi
         done
 
         [[ -n "${snooze}" ]] && [[ "${NotifyError}" == "false" ]] && update_snooze "${NotifyUpdates[@]}"
--- a/notify_templates/notify_v2.sh
+++ b/notify_templates/notify_v2.sh
@@ -189,9 +189,11 @@ dockcheck_notification() {
 
       if [[ ${#enabled_notify_channels[@]} -gt 0 ]]; then printf "\n"; fi
       for channel in "${enabled_notify_channels[@]}"; do
-        printf "Sending dockcheck update notification - ${channel}\n"
-        exec_if_exists_or_fail trigger_${channel}_notification || \
-        printf "Attempted to send notification to channel ${channel}, but the function was not found. Make sure notify_${channel}.sh is available in the ${ScriptWorkDir} directory or notify_templates subdirectory.\n"
+        if [[ ! "${channel}" == "file" ]]; then
+          printf "Sending dockcheck update notification - ${channel}\n"
+          exec_if_exists_or_fail trigger_${channel}_notification || \
+          printf "Attempted to send notification to channel ${channel}, but the function was not found. Make sure notify_${channel}.sh is available in the ${ScriptWorkDir} directory or notify_templates subdirectory.\n"
+        fi
       done
 
       [[ -n "${snooze}" ]] && [[ "${NotifyError}" == "false" ]] && update_snooze "dockcheck.sh"


yoyoma2:
The three new hardcoded checks for "file" in notify_v2.sh are my concern.

Current structure layers:

dockcheck.sh : simple hook to notify_v2.sh (or legacy v1)
notify_v2.sh : notification management features and dispatching
notify_xxx.sh : individual notification channel implementation details

Those "file" checks are bringing knowledge of implementation details up into the notify_v2 layer. If the management layer provides properties/capabilities and the individual notification channels pick and choose what they want, future maintenance nightmares will be avoided.

As usual just throwing out thoughts, feel free to ignore. This is a free script not a million dollar application and I'm not a merge cop so my thinking should be adjusted.

vorezal:
The idea of "channel properties" or similar is definitely what I was referencing when I mentioned a better, more flexible solution. I didn't like hard coding one channel to work significantly differently either. At the same time, having variables that set properties for every channel also seems a little bit too complicated. I'll think on it a bit more and see if there is some kind of middle ground we can come up with maybe.

vorezal:
I wonder if I've been too caught up in considering this as another notification channel and not its own thing. Maybe there should be another category called "integration channels" that only output or expose container update information for other tools or processes to consume? So instead of adding a lot of variables with properties for every notification channel, we just make a new list of integration channels (only file right now) which are designed to bypass snooze and update even when empty. This way it would be an addition with its own logic, have no impact on existing notification logic, and all we have to track in the config file is one additional list. This naming might be a little odd though considering it would still have to be handled by notify_v2.sh. Thoughts?

vorezal:
Gotcha. I was more thinking we could make another, broader category of notifications that just work the way this file one does. It would still reuse almost all the same code and run within notify_v2, but would only need one more list in the config file. Interesting thoughts around those use cases you mentioned though. Let me consider this a bit longer.

yoyoma2:
Got it. Your broader category idea doesn't sound like such a maintenance headache after all. The features you have in this PR for "file" are interesting and if you can somehow make them available to permit additional usecases that would be a bonus. Thanks for considering it.

mag37:
> @mag37 Just adding a warning when v1 is detected that legacy v1 is no longer tested and v2 should be used, would be a good start.

Yeah you're right @yoyoma2 , that should be easy enough. Though when v1 is removed, calling it v2 when it's the "only" way is a bit redundant, but changing that would break current functionality. So needs to be done with caution, or leave as is (the v2) while removing bits that's just there for v1.

---

Thank you  both for the brainstorming and discussing - I've had a hard time keeping up :smile: but I'm very grateful that you consider different possibilities / use cases and how to implement it most cleanly. 
I'm not entirely sure on the path forward with this - I don't know if I grasp all the bits you've been discussing either.

But let's say we've got our channels enabled like this:
- `NOTIFY_CHANNELS="DSM ntfy telegram"`

Couldn't we instead of having one (or multiple) extra variables for each notification setting have:
- `CHANNELS_BYPASS_SNOOZE="telegram"`
- `CHANNELS_BYPASS_DOCKCHECK_SELF="DSM telegram"`

etc.. So that all custom channel settings happen in one place instead of in every notification?   
If this is the path we'd go. Or is this what you mean @vorezal  with *"broader category of notifications"* ?

yoyoma2:
No rush to strip out v1. The new warning (use a noticeable color) to use v2 could be there for a year or more so if someone has a bug in v1 they will know to use v2.

@mag37 I very much like your idea of CHANNELS_BYPASS_SNOOZE centralized list type of variable if @vorezal can make that work. If someone has a "never forget" usecase channel that's already setting CHANNELS_BYPASS_SNOOZE and the "file" channel gets added they can't just add "file" to NOTIFY_CHANNELS and be done. The central location for these variables needs to be found and "file" added to each list variable that the "file" channel needs. A new version of "file" is released using a new list/variable,  every "file" user has to update their variables after upgrading.

Perhaps the individual non-standard channels call a function to add themselves to the lists they need.  Example bypass_snooze("file"), allow_empty_updates("file"), container_updates_only("file") so a new "file" channel release already adds itself to all necessary lists.

I could call container_updates_only("DSM") in my non-standard "containers only" usecase.

yoyoma2:
I like the feature list especially the template stuff. I will try to look at the code changes soon but I'm not too worried. Remember the jump from v1 to v2 looked daunting at first but it worked out well. I'll participate in testing too so confidence will be high before merging.

mag37:
I'll run some tests asap. Can do Telegram, Apprise and Ntfy for starters. Might have more channels ready.

mag37:
Good job correcting the errors!

:white_check_mark:   Telegram
:white_check_mark:   Apprise 
:ballot_box_with_check:   Ntfy - after a fix, see below.

Tested with and without Urls and Markdown options.


`notify_ntfy.sh` mistakenly does a check against Gotify-var:
```bash
# Wrong:
  if [[ -z "${!GotifyTokenVar:-}" ]] || [[ -z "${!GotifyUrlVar:-}" ]]; then
# Right:
  if [[ -z "${NTFY_DOMAIN:-}" ]] || [[ -z "${NTFY_TOPIC_NAME:-}" ]]; then
```

I've also got a minor suggestion - replacing the use of `tr` with bash parameter expansion.
Here's an example:
```bash
# use of tr:
    UpperChannel=$(tr '[:lower:]' '[:upper:]' <<< "$discord_channel")
# converted to parameter expansion:
    UpperChannel="${discord_channel^^}"
```
The same code is present in all tempaltes and multiple times in `..v2`.

yoyoma2:
I'm getting a garbled notification for the "file" template because there isn't an online version for it yet so the curl command fails as expected. Fixed by checking if LatestNotifySnippet is "undefined" (LatestNotifyRelease is never "undefined") which indicates a curl error. This was the minimal change to fix this problem but would also checking [[ -n $LatestNotifyRelease ]] make sense to detect other potential problems?
```
@@ -324,7 +324,7 @@ notify_update_notification() {
         RawNotifyUrl="https://raw.githubusercontent.com/mag37/dockcheck/main/notify_templates/notify_${NotifyScript}.sh"
         LatestNotifySnippet="$(curl ${CurlArgs} -r 0-150 "$RawNotifyUrl" || printf "undefined")"
         LatestNotifyRelease="$(echo "$LatestNotifySnippet" | sed -n "/${VersionVar}/s/${VersionVar}=//p" | tr -d '"')"
-        if [[ ! "${LatestNotifyRelease}" == "undefined" ]]; then
+        if [[ ! "${LatestNotifySnippet}" == "undefined" ]]; then
           if [[ "${!VersionVar}" != "${LatestNotifyRelease}" ]] ; then
             NotifyUpdates+=("${NotifyScript}.sh ${!VersionVar}  ->  ${LatestNotifyRelease}")
           fi
```

yoyoma2:
I tried DSM_ALLOWEMPTY=true and couldn't get it to work. This small _skip_snooze_ change fixes it but no idea if this breaks something else.
```
@@ -56,7 +56,7 @@ done
 skip_snooze() {
   local UpperChannel="${1^^}"
   local SkipSnoozeVar="${UpperChannel}_SKIPSNOOZE"
-  if [[ "${!SkipSnoozeVar:-}" == "true" ]]; then
+  if [[ -z "${snooze}" ]] || [[ "${!SkipSnoozeVar:-}" == "true" ]]; then
     printf "true"
   else
     printf "false"
```
It seems the following line won't allow DSM_ALLOWEMPTY=true to work when the snooze feature is not in use (SNOOZE_SECONDS is commented out). This is a line to which I added semicolons above.

`if { [[ $(skip_snooze "${channel}") == "true" ]] || [[ ${UnsnoozedContainers} -gt 0 ]] ; } && { { [[ "${MessageBody}" != "None" ]] && [[ "${MessageBody}" != '{"updates": []}' ]] ; } || [[ $(allow_empty "${channel}") == "true" ]] ; }; then`

vorezal:
Thanks as always for the thorough testing! I'll reevaluate my logic here and look at the fix for a missing online template file, but that last one is probably a very niche case that only developers will run into.

mag37:
Tested: ntfy + telegram, telegram configured with ALLOWEMPTY, CONTAINERSONLY, SKIPSNOOZE.

- `SKIPSNOOZE` and `ALLOWEMPTY` works fine! 
- Not getting dockcheck self notification no matter the configuration. 
- Not getting template update notification. (this is not even listed in the CLI)

yoyoma2:
Pulled the latest and dockcheck.sh and template notifications stopped working. Changed a couple logical ands to ors and they reappeared. Again I don't know what I'm doing here.

[and2or.txt](https://github.com/user-attachments/files/21992700/and2or.txt)

Oh and the semicolons are added correctly since DSM is happy now.

> but that last one is probably a very niche case

I don't know what you mean. You seem to have fixed the exact problem I reported but in a different way.

mag37:
I think it'd be sensible to break it up :) easier to read, easier to make sure the right logic is tested against the right values.
One-liners isnt worth hunting just by themselves.

vorezal:
I think I've got this refactored more fully. It seemed to make sense to factor the notification skip logic into a function now, and I noticed that if template notification variables were missing, the `return 1` actually caused a confusing error message to print out, so I suppose I'm considering this more of a `notification not sent successfully` case. I tested as many cases as I could on my end and didn't find any problems, but let me know if you come across anything I missed.

vorezal:
All good, and thanks! The readme probably does need some extra attention and simplification. Definitely let me know your thoughts. No rush though.

yoyoma2:
My testing of multiple channels from a same template worked but wasn't thorough. I'll try to test some more cases soon.
I thought the Readme was clear and didn't need to do any trial and error to make it do what I wanted. The only uncertainty was "any name as long as it is unique and contains no spaces or special characters". I read that as "any **alphanumeric** name as long as it is unique and contains no spaces or special characters" to be safe as I wasn't sure if an underscore is special.

vorezal:
I definitely think simplifying the readme and making it less of a wall-of-text would be great. Feel free to propose some adjustments, and I'll take a look if I have time as well. @yoyoma2 had a good point above about the wording around allowed notification channel names as well. Technically the characters allowed should be the exact same as Bash variable naming conventions, but those are a bit more nuanced than what I wrote. It should say that the name can include upper and lower case letters, numbers and underscores, but can't start with a number.

mag37:
I wrote a suggestion of the readme-changes yesterday though didn't wanna shove it into your PR without your eyes on it (both @yoyoma2 and @vorezal ).

I'll paste it in another comment so it can be previewed.

mag37:
Suggestion of a new Notifications-section:
___
___

## :loudspeaker: Notifications
Triggered with the `-i` flag. Will send a list of containers with updates available and a notification when `dockcheck.sh` itself has an update.  
`notify_templates/notify_v2.sh` is the default notification wrapper, but if a `notify.sh` is present and configure it will override.  

Example of a cron scheduled job running non-interactive at 10'oclock excluding 1 container and sending notifications:  
`0 10 * * * /home/user123/.local/bin/dockcheck.sh -nix 10 -e excluded_container1`

#### Installation and configuration:
Set up a directory structure as below.  
You only need the `notify_templates/notify_v2.sh` file and any notification templates you wish to enable, but there is no harm in having all of them present.
```
 .
├── notify_templates/
│   ├── notify_DSM.sh
│   ├── notify_apprise.sh
│   ├── notify_discord.sh
│   ├── notify_generic.sh
│   ├── notify_gotify.sh
│   ├── notify_HA.sh
│   ├── notify_matrix.sh
│   ├── notify_ntfy.sh
│   ├── notify_pushbullet.sh
│   ├── notify_pushover.sh
│   ├── notify_slack.sh
│   ├── notify_smtp.sh
│   ├── notify_telegram.sh
│   └── notify_v2.sh
├── dockcheck.config
├── dockcheck.sh
└── urls.list         # optional
```
- Uncomment and set the `NOTIFY_CHANNELS=""` environment variable in `dockcheck.config` to a space separated string of your desired notification channels to enable.
- Uncomment and set the environment variables related to the enabled notification channels. Eg. `GOTIFY_DOMAIN=""` + `GOTIFY_TOKEN=""`.  

It's recommended to only do configuration with variables within `dockcheck.config` and not modify `notify_templates/notify_X.sh` directly.
If you wish to customize the notify templates yourself, you may copy them to your project root directory alongside the main `dockcheck.sh` (where they're also ignored by git).
Customizing `notify_v2.sh` is handled the same as customizing the templates, but it must be renamed to `notify.sh` within the `dockcheck.sh` root directory.


#### Snooze feature:
Configure to receive scheduled notifications only if they're new since the last notification - within a set time frame.  

**Example:** *Dockcheck is scheduled to run every hour. You will receive an update notification within an hour of availability.*  
**Snooze enabled:** You will not receive a repeated notification about an already notified update within the snooze duraion.  
**Snooze disabled:** You will receive additional (possibly repeated) notifications every hour.  

To enable snooze uncomment the `SNOOZE_SECONDS` variable in your `dockcheck.config` and set it to the number of seconds you wish to prevent duplicate alerts.  
Snooze is split into three categories; container updates, `dockcheck.sh` self updates and notification template updates.  

If an update becomes available for an item that is not snoozed, notifications will be sent and include all available updates for that item's category, even snoozed items.  

The actual snooze duration will be 60 seconds less than `SNOOZE_SECONDS` to account for minor scheduling or run time issues.  


#### Current notify templates:
- Synology [DSM](https://www.synology.com/en-global/dsm)
- Email with [mSMTP](https://wiki.debian.org/msmtp) (or deprecated alternative [sSMTP](https://wiki.debian.org/sSMTP))
- Apprise (with it's [multitude](https://github.com/caronc/apprise#supported-notifications) of notifications)
  - both native [caronc/apprise](https://github.com/caronc/apprise) and the standalone [linuxserver/docker-apprise-api](https://github.com/linuxserver/docker-apprise-api)
  - Read the [QuickStart](extras/apprise_quickstart.md)
- [ntfy](https://ntfy.sh/) - HTTP-based pub-sub notifications.
- [Gotify](https://gotify.net/) - a simple server for sending and receiving messages.
- [Home Assistant](https://www.home-assistant.io/integrations/notify/) - Connection to the notify [integrations](https://www.home-assistant.io/integrations/#notifications).
- [Pushbullet](https://www.pushbullet.com/) - connecting different devices with cross-platform features.
- [Telegram](https://telegram.org/) - Telegram chat API.
- [Matrix-Synapse](https://github.com/element-hq/synapse) - [Matrix](https://matrix.org/), open, secure, decentralised communication.
- [Pushover](https://pushover.net/) - Simple Notifications (to your phone, wearables, desktops)
- [Discord](https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks) - Discord webhooks.
- [Slack](https://api.slack.com/tutorials/tracks/posting-messages-with-curl) - Slack curl api

Further additions are welcome - suggestions or PRs!  
<sub><sup>Initiated and first contributed by [yoyoma2](https://github.com/yoyoma2).</sup></sub>

#### Notification channel configuration:
All required environment variables for each notification channel are provided in the default.config file as comments and must be uncommented and modified for your requirements.
For advanced users, additional functionality is available via custom configurations and environment variables.
Use cases - all configured in `dockcheck.config`:  
(replace `<channel>` with the upper case name of the of the channel as listed in `NOTIFY_CHANNELS` variable, eg `TELEGRAM_SKIPSNOOZE`)  
- To bypass the snooze feature, even when enabled, add the variable `<channel>_SKIPSNOOZE` and set it to `true`.
- To configure the channel to only send container update notifications, add the variable `<channel>_CONTAINERSONLY` and set it to `true`.
- To send notifications even when there are no updates available, add the variable `<channel>_ALLOWEMPTY`  and set it to `true`.
- To use another notification output format, add the variable `<channel>_OUTPUT` and set it to `csv`, `json`, or `text`. If unset or set to an invalid value, defaults to `text`.
- To send multiple notifications using the same notification template:
  - Strings in the `NOTIFY_CHANNELS` list are now treated as unique names and do not necessarily refer to the notification template that will be called, though they do by default.
  - Add another notification channel to `NOTIFY_CHANNELS` in `dockcheck.config`. The name can contain upper and lower case letters, numbers and underscores, but can't start with a number.
  - Add the variable `<channel>_TEMPLATE` to `dockcheck.config` where `<channel>` is the name of the channel added above and set the value to an available notification template script (`slack`, `apprise`, `gotify`, etc.)
  - Add all other environment variables required for the chosen template to function with `<channel>` in upper case as the prefix rather than the template name.
    - For example, if `<channel>` is `mynotification` and the template configured is `slack`, you would need to set `MYNOTIFICATION_CHANNEL_ID` and `MYNOTIFICATION_ACCESS_TOKEN`.

### :date: Release notes addon
There's a function to use a lookup-file to add release note URL's to the notification message.  
Copy the notify_templates/`urls.list` file to the script directory, it will be used automatically if it's there.  
Modify it as necessary, the names of interest in the left column needs to match your container names.  
To also list the URL's in the CLI output (choose containers list) use the `-I` option or variable config.  
For Markdown formatting also add the `-M` option. (**this requires the template to be compatible - see gotify for example**)  

The output of the notification will look something like this:
```
Containers on hostname with updates available:
apprise-api  ->  https://github.com/linuxserver/docker-apprise-api/releases
homer  ->  https://github.com/bastienwirtz/homer/releases
nginx  ->  https://github.com/docker-library/official-images/blob/master/library/nginx
...
```
The `urls.list` file is just an example and I'd gladly see that people contribute back when they add their preferred URLs to their lists.

vorezal:
Yeah that looks slightly off. I would say "if notify.sh is present and configured, it will override" I'll read through in more detail later today though.

As for credit, I say leave it. Then @yoyoma2 is responsible for future maintenance and changes 🎉

vorezal:
Sure I can put the changes in shortly along with anything else I may find. I think otherwise we're looking good. At least so far. Wider testing might be needed to know for certain across all notification channels though I suppose.

mag37:
Tested a few runs with different setups and different channels, pushed to main.

Great job, thank you for the hard work @vorezal and valuable input and testing @yoyoma2  !
Did some minor tweaks and made a new release - hopefully we wont hit any severe bumps.

PR #235, opened 2025-10-29
Add -R flag to skip container recreation after pulling images

napalmz:
#### Summary
This PR introduces a new optional `-R` flag.  
When used, the script pulls updated images but **skips the container recreation phase**.  
This is useful for CI pipelines, staged rollouts, or manual maintenance windows.

---

#### Changes
- **Help**: Added documentation for the `-R` option.  
- **Initialization**: Added default variable `SkipRecreate=false`.  
- **Argument parsing**: Extended `getopts` to handle the `-R` flag.  
- **Execution flow**: After pulling images, the script now checks `SkipRecreate` and exits early if enabled, while still allowing optional `docker image prune`.

---

#### Usage
```bash
# Pull updates only, do not recreate containers
./dockcheck.sh -R

# Combine with auto mode
./dockcheck.sh -a -R

# Optionally prune dangling images even when skipping recreation
./dockcheck.sh -a -R -p
```

### Behavior
- Pulls updated images as usual.  
- If `-R` is provided:
  - Skips the container recreation step (`docker compose up`, `docker run`, etc.).
  - Optionally performs `docker image prune` if `-p` is set or confirmed by the user.
- When `-R` is not used, the script behaves exactly as before.

---

### Motivation
- Separate image pulling from container recreation for better control.
- Useful for automated pipelines, scheduled maintenance, or multi-stage deployments.
- Reduces downtime by allowing bulk recreations later.

---

### Backward compatibility
No breaking changes introduced.  
If `-R` is omitted, the script executes the full standard workflow.

---

### Test plan
1. Run with `-R` → verify images are pulled but containers are **not** recreated.  
2. Run with `-a -R -p` → verify images are pulled and prune executes.  
3. Run without `-R` → verify normal recreate occurs after pull.  
4. Run with invalid flags → confirm existing argument parsing still works as expected.

---

### Notes
- Compatible with both `docker compose` and legacy `docker run` logic.  
- Can be combined with `-a`, `-p`, or other flags.  
- Output clearly indicates when recreation is being skipped.

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -557,60 +557,57 @@ if [[ -n "${GotUpdates:-}" ]]; then
 
       docker pull "$ContImage" || { printf "\n%bDocker error, exiting!%b\n" "$c_red" "$c_reset" ; exit 1; }
     done
-    printf "\n%bDone pulling updates. %bRecreating updated containers.%b\n" "$c_green" "$c_blue" "$c_reset"
+    printf "\n%bDone pulling updates.%b \n" "$c_green" "$c_reset"
+    
     if [[ "$SkipRecreate" == true ]]; then
-      # Optional prune even when skipping recreation
-      if [[ "$AutoPrune" == false ]] && [[ "$AutoMode" == false ]]; then printf "\n"; read -rep "Would you like to prune all dangling images? y/[n]: " AutoPrune; fi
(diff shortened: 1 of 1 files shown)


napalmz:
Hello @mag37 
Your solution is definitely fine.
All I need is to be able to skip rebuilding the containers.

Thank you and have a nice day!

PR #239, opened 2025-11-12
Fix version check condition

astappiev:
Happens to me had a misconfigured DNS, which resulted in a bug where `dockcheck.sh` always detects a new version and replaces itself with a blank file.

A quick investigation showed that this snippet never produces `"$LatestRelease" == "undefined"` and instead `undefined` is only assigned to `"$LatestSnippet"`
```bash
# Check if there's a new release of the script
LatestSnippet="$(curl ${CurlArgs} -r 0-200 "$RawUrl" || printf "undefined")"
LatestRelease="$(echo "${LatestSnippet}" | sed -n "/VERSION/s/VERSION=//p" | tr -d '"')"
LatestChanges="$(echo "${LatestSnippet}" | sed -n "/ChangeNotes/s/# ChangeNotes: //p")"
```

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -358,7 +358,7 @@ list_options() {
 }
 
 # Version check & initiate self update
-if [[ "$LatestRelease" != "undefined" ]]; then
+if [[ "$LatestSnippet" != "undefined" ]]; then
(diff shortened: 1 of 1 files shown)


mag37:
Nice catch! 

Sloppy of me, been meaning to look at some more error handling soon :tm:

PR #261, opened 2026-02-02
Docker configuration

### LINKED ISSUE #77: Idea: package the repo into an container
Hi there!

Just an idea: why not package the script / repo contents into a container. Then we can use it to run everything inside docker - no need to install dependencies and stuff locally.

* docker socket needs to be mounted in the container
* docker compose needs to be installed (alongside other requirements of this repo)
  * maybe the container can be based on docker/compose image?

What do you think of that?
Best wishes,
koseduhemak

vorezal:
I've seen a few requests to run Dockcheck within Docker. This could use a lot more testing than I was able to do with only my use cases, but here is what it should provide:

- Dockerfile for image building. Anyone can use this to build one themselves, and we can also publish it in Dockerhub. I think it would make the most sense to do this under the mag37/dockcheck namespace, but I would be happy to do it under vorezal/dockcheck and maintain it if that's easier.
- Example compose.yaml file which can be modified for individual use cases.
- Manages schedules and command line arguments as environment variables passed to the container.
- Documentation not yet written. I'd like feedback and testing to confirm it is fully functional, then I'll write up documentation.
- I think the container can even update itself, but it could short circuit an execution and I still need to test that case.

Limitations I can think of at present:

- DSM notification won't work if running on DSM as a container
- The files snooze.list and updates_available.txt won't live beyond the container unless added to a volume. This may be fine as is, or may need additional documentation and examples.
- Apprise limited to curl configuration unless we add the package to the container image, which would increase its size.
- Rootless docker installation not supported and this container runs as root to access docker.sock. I think this is ok since privilege escalation is pretty trivial for anything that can run arbitrary containers.
- Compose projects path must exactly match host system path and every compose file must be added to a volume to be discovered.

As always, open to thoughts and improvements!


mag37:
Ah then I get what you mean! So let's say I've got this layout (for some reason):
```sh
/opt/projects/containers/prod
├── container_one
│   ├── data
│   └── docker-compose.yml
└── container_two
     └── docker-compose.yml


/home/user_bob/containers/test
├── container_t1
│   ├── data
│   └── docker-compose.yml
└── container_t2
     └── docker-compose.yml
```

I'd need to bind both of them with:
```yml
    volumes:
      - /home/user_bob/containers/test:/home/user_bob/containers/test:ro
      - /opt/projects/containers/prod:/opt/projects/containers/prod:ro
 ```
Or as you state, bind more granular with individual containers etc.

___

>namespace

I'd be fine running it under `mag37/dockcheck` as long as you think it's fair/sensible. I'll probably stay away from using Github actions for now and rather build+publish manually on new releases.

(I'm considering if a migration to Codeberg is sensible in the future - though that's a later discussion/project when time allows)

**Comment by vorezal** [comment]
Yes that volume structure looks correct! And sounds good on the docker image publishing. I think keeping mag37/dockcheck is quite sensible and the level of effort is pretty minimal.

**Comment by vorezal** [comment]
Ah! I completely forgot that urls.list would become required when added to the build. We can add that as a volume (optional) in the example docker compose file as well. Adding the container_name and hostname also makes sense. If we are taking on building and publishing the image though, I'm not sure we need the build directive. Unless we expect people to build a local image themselves. Which is also fine really, but I was writing it as if the user would pull from the dockerhub repository.

I can take care of the urls.list fix and add some comments to the compose-example.yaml file. I'll hold off on the container_name, hostname, and build directives in case you want to talk about those a little more.

**Comment by mag37** [comment]
> I can't think of a way to counter that, at least on our side. We could certainly call out the security risk and recommend that users install Docker in rootless mode, but that would be a host management thing.

I agree, from what I can think of it's hard to work around as those levels of access is needed to do the docker handling. 
Though I think we should write a warning/disclaimer of some sort - at least in the readme but maybe also in the compose file.
Maybe if users only want to use notifications they can have a stricter socket permission. Sadly mounting the `docker.sock` as `:ro` only makes the user unable to manipulate the device (which is probably sensible anyway but might also lead to a false sense of security) but does not affect actual docker permissions. To adjust that it has to be run with a socket proxy of some sort which will block specific API calls.

> I think I still like the idea of building the script files into the container and publishing it ourselves so it is fully encapsulated and the version of the container matches the version of the script within it. I do think the chance of an image takeover is quite low, or at least about the same as Github getting taken over.

I agree - the security risk is probably still the same either way and the alternative is not as clean as including everything needed within the container.  

It's a hard thing to swallow though, the security risk. But at the same time - running it "bare metal" currently is as much of a risk anyway so it probably just that we need to be clear that it's more or less host level access that people can mitigate by using rootless docker and other things manually.

**Comment by jadero** [comment]
Very nice. I would use this on my LibreELEC instances since this doesn't come with bash, but instead with busybox ash.

**Comment by mag37** [comment]
I just ran this through a few tests again and I really think I could finalize it and merge it asap.
The comments in the compose file should be enough, I might rephrase/refine some and then just make a brief section in the README too. 

Then comes the step of providing built and ready images - I'll need to look into how to do that for multiple architectures.
I've registered this at least: https://hub.docker.com/r/mag37/dockcheck

**Edit:** Ok so I think I've got a local build-env setup that can produce different images for multiple architectures and push to dockerhub. Though we'd probably need to test on some but I think we should aim to have `linux/amd64` and `linux/arm64` to start with.

**Comment by mag37** [comment]
I started pushing edits straight into your branch @vorezal I hope you dont mind.

There's a few images up in the registry at https://hub.docker.com/r/mag37/dockcheck now, I build one for last v0.6 and v0.7 for some testing+history. I've decided I'll build 3 tags each release; *major*, *minor* and *latest* (which ofc only sticks to the latest release) - so **v0.8.1** would be: `:v0.8` , `:v0.8.1` and `:latest` when it gets released.  
Does this sounds reasonable?

I've also settled on these architectures: 
- linux/amd64
- linux/arm64
- linux/arm/v7

Though I've got to setup some testing environments for the arm platforms to see its working properly. Or find volunteers who has that in "production" :)

I'm thinking this is pretty much ready to merge as it does not affect anything outside the readme.

**Comment by mag37** [comment]
Due to a question/suggestion by a user I was briefly testing to use env vars instead of the `dockcheck.config` and that seems to work fine too!

Like:
```yml
services:
  dockcheck:
    container_name: dockcheck
    hostname: dockerHostXYZ # The name that will be used in notifications
    image: mag37/dockcheck:latest
    restart: unless-stopped
    environment:
      NOTIFY_CHANNELS: "xmpp"
      XMPP_SOURCE_JID: "abc"
      XMPP_SOURCE_PWD: "123"
      XMPP_DEST_JID: "xyz"
      BackupForDays: "5"
      CRON_SCHEDULE: "0 */3 * * *"
      DOCKCHECK_ARGS: "-mniI -x 10 -e container1"
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - /var/run/docker.sock:/var/run/docker.sock # Be aware that this is insecure, container has full control
      - /path/to/projects/docker:/path/to/projects/docker:ro  
```


I'll need to do some more testing with that - could be worth adding to the docs if so.

vorezal:
Thanks for looking at this again!

- No issue at all pushing change straight to my branch. Feel free.
- Overall everything sounds pretty reasonable. I wouldn't have thought to target ARMv7, but if we want to support older Raspberry Pi devices that could be useful. Maybe ARMv8 as well for newer versions? I'll admit though I am very unfamiliar with these.
- Setting environment variables within the Docker compose file will absolutely work exactly the same as specifying them in dockcheck.config and is a valid option or alternative. Documenting that makes great sense. I'm also pretty sure that if the same option were specified in both places dockcheck.config would take precedence. I'd want to test that before documenting it though.

mag37:
Thank you, I'll touch up the readme some more and the docker hub info page. Then I'll push this to main.

ARMv7 is indeed due to some older rPI models and the ARMv8 is included in the arm64 build.

Yeah that's great with the env vars! I didn't think of that before but a user pointed out that it might be simpler for some to just have everything in the compose with env vars than having a config-file. Reasonable.
I just tested and you're correct - the `dockcheck.config` takes precedence.

Thank you for the work, again! :rocket:

vorezal:
Ah! I didn't realize ARMv8 and ARM64 were effectively the same thing. And thanks for testing!

mag37:
I just realized something I forgot to account for (while I did some testing).
@vorezal  did you test some way of handling self-updates from within the container?

Currently (if I remember correctly from my test) I think it will run through pulling the newer image and then when recreating it will crash (understandably, as it's restarting) and then fail to start.

vorezal:
I have not tested a self update yet. I was expecting it to interrupt the process when it updates itself, but I'm not sure it will fail to start or not.

mag37:
I'll need to do some further testing and think of the alternatives other than telling users to be careful and maybe consider adding `-e dockcheck` if they run unattended automated updates. 

I need to test both letting the container update itself as a container. And let the containerized script update itself as a script to see what results we've got to handle.

Otherwise it seems like a success this far, managed to get some tests on both a ARM64 and a ARMv7 today which both worked good.

mag37:
Script self update works while its running, but that'll "reset" if it stops+starts (as its not persistent) and will still find docker updates for itself. 

Docker self updates crashes the container, it stops it but never starts it properly again. 

I'll try to look deeper when I've got time, but either update path is currently not working properly - only a regular `docker compose pull && docker compose up -d` or similar on the host will work. 

Need to think about alternatives. 


**Edit**: created an issue where we'll continue the discussion.

Issue #260, opened 2026-02-01
when one container update fails, all other tasks are stopped/failed/unfinished


underwoodblog:
As the title said, dockcheck.sh -p -u found multiple containers for update that where not related to each other. The last one failed (timeout of big slow download?) while downloading and all other failed too.  worked when running each container update alone.

Issue #249, opened 2026-01-07
Send a notification with the whole run


MohandL3G:
Sorry if this has been requested before tried to look for it but couldn't find it, I run dockechk with -api and I get a message on discord notify me for containers with updates like this:

```
🐋 Containers on mohandl3g-server with updates available:
n8n
```

This is cool, I also want it to send another notification after the whole run, so I know if it has been updated or an issue or something happened, and also if the autoprune finished, something like this for example:

```
Containers Updated:
n8n

- Auto Prune has removed 1.2GB
```

Sorry english is not my main and thanks for the hard work

mag37:
> Seems reasonable. I'll see if I can take a look at some point.

You don't **have to**! But if you want to, there's no rush. I won't have time to spare for another couple of weeks most likely.

vorezal:
No worries! Happy to do it. Time permitting of course.

yoyoma2:
> It have been mentioned as an idea before I think - as an idea at least.
> 
I was trying to remember where I had heard about this before. Now I recall a discussion where a DSM user was enabling a DSM scheduled task feature that sends you the entire output and saying it's up up to user preference if they prefer the full output or the dockcheck produced notifications.

Seems a bit redundant to me. If a pending update hung around at each dockcheck run I'd investigate but dockcheck just works. No objection but hopefully any new notification can be turned off if not desired.

mag37:
I was to point out the PR in progress but saw your edit @yoyoma2 :) 

I think your input is good though, and maybe more suited for a "debug"-suggestion in the readme or like a future debug-wrapperscript or something. As I think the OP of this issue was aiming more at getting the transactional log of what got updated and what/if some failed - which hopefully will be solved by the #262 in the future!

Issue #244, opened 2025-12-13
Idea: create timestamped logfile for each run


dlong500:
I see the new option to create backups of images, but it would be nice to have the option to save a timestamped log of activity per run that would include the **tag and digest string of the old version** (along with the new version).  This would avoid having to take up the space that actual image backups need but still make it fairly easy to roll back to an older version via the image digest if necessary.

mag37:
Ah yes, that should be possible. Just need to grab those values as in the POC pre-pull and then grab the new digest post-pull with something like:
```bash
NewDigest=$(docker image inspect "$ContImage" --format '{{index .RepoDigests 0 }}')
```

I'll look into creating a branch for it when I got time!

PR #262, opened 2026-02-03
Start/restart stacks once. Skip failed pulls. Final summary notification.

vorezal:
Attempting to address issues #249 and #260. Additionally, I skipped running `docker compose up` against stacks that it had already been performed for. Open to thoughts on that last bit, since no one actually asked for it but me. 

I've tested a fair amount, but I don't think I hit every case. I wasn't able to "successfully" cause a docker pull to error or even think of how to force it to do so as a test. I'll keep testing more myself, but I wanted to get this up for review/thoughts before I went too far with a particular approach that someone else may know of a better way to handle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,8 @@ regctl
 snooze.list
 # ignore updates file
 updates_available.txt
+# ignore user compose files
+compose.yaml
+compose.yml
+docker-compose.yaml
+docker-compose.yml
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ compose.yaml
 compose.yml
 docker-compose.yaml
 docker-compose.yml
+# ignore log directory
+/log/
--- a/default.config
+++ b/default.config
@@ -44,6 +44,8 @@
 # DISABLE_DOCKCHECK_NOTIFICATION=false
 ## Uncomment and set to true to disable notifications when notify scripts themselves have updates.
 # DISABLE_NOTIFY_NOTIFICATION=false
+## Uncomment and set to true to enable a final action summary notification.
+# ENABLE_SUMMARY_NOTIFICATION=false
 #
 ## Apprise configuration variables. Set APPRISE_PAYLOAD to make a CLI call or set APPRISE_URL to make an API request instead.
 # APPRISE_PAYLOAD='mailto://myemail:mypass@gmail.com
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -13,20 +13,6 @@ ScriptArgs=( "$@" )
 ScriptPath="$(readlink -f "$0")"
 ScriptWorkDir="$(dirname "$ScriptPath")"
 
-# Source helper functions
-source_if_exists_or_fail() {
-  if [[ -s "$1" ]]; then
-    source "$1"
-    [[ "${DisplaySourcedFiles:-false}" == true ]] && echo " * sourced config: ${1}"
-    return 0
-  else
-    return 1
-  fi
-}
-
-# User customizable defaults
-source_if_exists_or_fail "${HOME}/.config/dockcheck.config" || source_if_exists_or_fail "${ScriptWorkDir}/dockcheck.config"
-
 # Help Function
 Help() {
   echo "Syntax:     dockcheck.sh [OPTION] [comma separated names to include]"
@@ -49,6 +35,7 @@ Help() {
   echo "-M     Prints custom releasenote urls as markdown (requires template support)."
   echo "-n     No updates; only checking availability without interaction."
   echo "-p     Auto-prune dangling images after update."
+  echo "-q     Quiet mode. Minmal text output. Does not effect file based logging, if enabled."
   echo "-r     Allow checking/updating images created by 'docker run', containers need to be recreated manually."
   echo "-R     Skip container recreation after pulling images."
   echo "-s     Include stopped containers in the check. (Logic: docker ps -a)."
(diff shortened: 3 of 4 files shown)


vorezal:
Agreed. My first pass at that was pretty rough. I'll try to clean up the logic. My IDE also kept auto-indenting in a way that I didn't intend to and I didn't put in the effort to fix it at the time. I'll clean that up too.

vorezal:
That's a good thought. Building a logging system once within dockcheck and then having notification logic to parse through for log message types or verbosity levels would be more work, but it would be more segregated and potentially more maintainable. Let me think about how this could be done.

Apologies, I forgot to address that source change. At one point I was troubleshooting an error causing the source to fail. That was meant as a troubleshooting line and forgetting to remove it was a miss on my part. I'll make sure to get it out of there.

vorezal:
I have some ideas on this and have actually already been working on it off and on. I should have something ready to show in a little bit. Some of the basic features I'm considering and would be happy to have some initial feedback on:

- A log-and-print function that writes to the screen, stores the message in a memory buffer, and (optionally) logs to file.
- A log only function that just stores in a memory buffer and logs to file
- Log levels based on the OpenTelemetry log data-model severities: [link to docs](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber)
- A function that writes a stored buffer out to the screen and/or triggers a notification. This would end up here and there throughout the dockcheck.sh code (really just one place to start) but I can't really think of a way around that. One way or another we need something in the main script to trigger an alert.
- Wrapper for function execution that can control stdout output while maintaining return code
- Quiet mode flag -q to prevent output to stdout
- Config option for log level output
- Config option to enable file based logs
- Config option for log file retention

I think those are the main points. If that sounds reasonable I'll get the functions finished.

vorezal:
Alright folks. Long comment (and PR). Sorry for that.

I've added several functions to support standardized logging. It requires more additional characters in code than I might like. I'd love for it to be as simple as calling a function with a message and have that be it, but to support buffers and log levels I don't see a way to avoid requiring at least those as parameters. I suppose separate functions for each log level and with/without buffer storage could work now that I think about it, but could be too much. Open to thoughts there however. Anyway, I also changed a couple of lines to use the new log functions as examples, but left most unchanged for now. I expect functionality be as follows:

`log_print buffername INFO "Some message\n"` <- Prints the message to the screen, stores it in the buffer buffername, and may write out to file if configured in the settings
`log buffername INFO "Some other message\n"` <- Stores the log in mainbuf and may write out to file if configured, as above. Does not print to stdout.
`print_buffer buffername [notify/notifyonly] [label]` <- Prints the buffer. Parameters 2 and 3 are optional, with the default being to print to stdout only. If 'notify' is passed, it will also send a buffer notification. If 'notifyonly' is passed, it will only send the notification without printing to stdout. If notify or notifyonly are passed and parameter 3 is sent, its value will be used in place of the buffer name as the label for the notification. This will need to be added to dockcheck.sh wherever/whenever we want a new buffer printed out or a new buffer notification (if we ever make another).
`log_severity string` <- Takes a string and converts it to a numerical representation of the log severity. If a number is passed between 0 and 24, uses it directly. If anything else is passed, defaults to 0. Just a helper function.
`log_cleanup` <- Removes log files older than LogDaysToKeep days.

`exec_command command and arguments` <- Takes a normal bash command as arguments, runs it, stores the output in a variable, and returns the response code. If Quiet is not set, also allows printing to stdout via a duplicated file descriptor
`LatestOutput` <- Variable used to store the output from the last executed command sent through exec_command. We can use this to dynamically push the response to a buffer and/or log it to file.

A few points about logging and my assumptions:

- I assume we handle newlines in stdout manually. No newlines are added automatically to screen output.
- I think we should only log single lines to file. I took the easy way out for now and replace every newline with a literal '\n' and add a single newline to the end. We could alternatively log in json or something, but this seemed the least complicated. We could even decide to support multi-line log events, but that's not generally considered best practice these days.
- Logs are kept in the `/log` directory within the main project folder and is created if it does not exist.
- A new log file is created for each exeution.
- I expect strings in buffers we notify on to have `;` characters separating up to three fields. The action or event name, optional status of the action or event (success, error, skipped, whatever is appropriate for that event), and an optional detailed message or description. Only the first field is required, but the `;` separator, when provided, does make file logs look a little weird. I haven't decided what, if anything, to do about that for now.

New parameters:
`-q` : Quiet

New configuration options:
`LogToFile`: true/false
`LogDaysToKeep`: integer
`LOG_LEVEL`: INFO/WARN/etc
`ENABLE_SUMMARY_NOTIFICATION`: true/false

README changes: TBD once this is reviewed and more finalized, but for the most part we won't need to document the internal log changes. Just the new parameter and configuration options.

I expect a lot of testing to be needed here. Even more so when we start migrating log messages over. I'm almost positive we will find interesting formatting cases I hadn't accounted for that makes things come through differently or unexpectedly and we will have to adjust. Some of this may well be... over engineered at best.

Finally, I rewrote the logic to check if a stack has already been restarted. It's more verbose with full multi-line if statements, but it should hopefully be easier to follow. Still might need additional comments for clarity though.

Phew. That was a bunch. Let's split this out and test each feature change one by one.

vorezal:
Yeah this was a combination of things. One was to allow the capturing of the output at all (sending it to a buffer if needed). The other though was to handle the quiet flag easily, but that could be done other ways if docker CLI output is troublesome. The main issue seems to be that when you do anything  to the Docker CLI output I think it detects it isn't part of an interactive shell and defaults to text only output, line by line. I think I might have a way of adjusting for that, but if not we may need to handle Docker output, or any other output that relies on a TTY, separately without having a logger for it.

vorezal:
Yeah I think you're right here. I may have gone overboard on this one unnecessarily, and there are probably a few too many edge cases and issues to handle them all properly with my attempt at an execute_command wrapper function. We'll likely need to capture stdout and handle logging/printing to screen separately in each case. I think it should still be very reasonably possible to handle the quiet flag anyway, but I was concerned about the dynamic progress output interfering. I think we can manage that by adopting the Docker practice (if we want to) of sending progress bars and the like to stderr instead of stdout though.

mag37:
I'm considering if we should re-think this a bit before you plow down too much time to rewrite anything :)
Maybe break things down a bit and even split it to different things.

I'm not writing this to control or force anything, I'm just afraid things snowball to be a much larger scope and perhaps harder to maintain than it needs to be. And @vorezal crushing themselves with work! 
So read it with a tone of good faith.

---

- Do we need the quiet-flag? 

A manual (and blunt) way is to just append `&>/dev/null` to have it completely quiet. What's the use case we're aiming for here? If this implementation is what causes the most rewrites or issues maybe the worth should be reconsidered.

- What is necessary to fulfill the logging scope? 

What's suggested and asked for was the ability to get a full summary of what's successfully updated to be used within notifications. I was also looking at keeping a logfile with the digests as a transaction log, but not completed it yet (see branch: [log_digests](https://github.com/mag37/dockcheck/tree/log_digests)) as suggested in [#244](https://github.com/mag37/dockcheck/issues/244).

- Another issue was [#260](https://github.com/mag37/dockcheck/issues/260) - to continue the run even if an update fail.

We should handle the fails gracefully and keep track of succeeded//failed updates to be able to skip recreation and maybe include easier in notifications.

- Skip running `docker compose up` against stacks that had already been recreated.

This is very sensible and would make things smoother in general, so I think this is very much worth it.

vorezal:
I think very reasonable in this case considering how different the features are. I'll work on pulling them into separate branches and we can evaluate them individually.

PR #267, opened 2026-02-28
Allow ranges to be used for selection

max-foss:
Hi all,

I built this small feature for myself and wanted to share it with the rest of you.

`Choose what containers to update.
Enter number(s) separated by comma or range(s) (e.g. 1-3,4-5), [a] for all - [q] to quit: 2-22
`

Let me know if I need to adjust anything before it can be merged.

Cheers

Max

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Containers with updates available:
 03) whoogle-search
 
 Choose what containers to update:
-Enter number(s) separated by comma or range(s) (e.g. 1-2,4-5), [a] for all - [q] to quit: 1-2
+Enter number(s) or range(s) separated by comma (e.g. 1-2,4-5,09), [a] for all - [q] to quit: 1-2
 ```
 
 Then it proceeds to run `pull` and `up -d` on every container with updates.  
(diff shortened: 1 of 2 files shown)


mag37:
Nice addition, nice use of `BASH_REMATCH` too! Had forgot about that one so had to look it up :) 

I did tests with combinations like `1-3,5,6-09` and intentional breakage like `-1-4` and out of range. Seems to hold up!

max-foss:
Now I remember what this reminded me of, the custom multi page print dialog found in any halfway decent application or print dialog:
Bildschirmfoto 2026-03-01 um 10 23 16

mag37:
> > Nice addition, nice use of `BASH_REMATCH` too! Had forgot about that one so had to look it up :)
> 
> Hmm, should we add an additional comment to explain that?


I don't think thats necessary, its standing out and is an easy search - while being a bit long to explain as a concept in a brief comment.

mag37:
Yeah thats no worries though, you can try pulling a image 100 times and it will already have it pulled. 

And keeping track of what already updated/recreated is being worked upon elsewhere. A bit of a larger scope. Thanks though, this PR is fine as is 😃

PR #271, opened 2026-03-12
Docker restart/pull handling updates. Minor refactoring.

### LINKED ISSUE #260: when one container update fails, all other tasks are stopped/failed/unfinished

vorezal:
This PR:

- Skips restarting already restarted stacks
- Skips failed image pulls
- Moves some stuff around/minor refactoring to allow other bash functions to be used in the print_backups function

This is the first PR in my attempt to separate features in #262. A log feature only PR will follow eventually. This one is intended to address #260.


mag37:
Sorry @vorezal - other things sprung up in between (as you know) and time has been scarce.

I just pushed another release to have one fresh before merging this. 
I'll work out the merge conflicts (dosnt look like much) and do some testing locally, so nothing else affected this change, before I merge it to main. I might look into squeezing in some more cleaning and bump it to `v0.8.0` to mark a new chapter with this :smile:

vorezal:
Thanks! Sounds good, and no worries. Time is definitely scarce all around.

mag37:
Epic work as always @vorezal  - the conflicts were minor, should've done this earlier but just needed the headspace to feel I grasped things for testing etc. 

Ran it through a few tests with containers in different states and stacks. Thank you again.

vorezal:
Awesome. Thanks for merging!

PR #279, opened 2026-04-11
fix: avoid printf escape expansion in dockcheck update body

### LINKED ISSUE #278: Self-update notification produces invalid JSON when using the file channel
When dockcheck.config:
```
NOTIFY_CHANNELS="file"
FILE_OUTPUT="json"
```
If an update of dockcheck.sh is found, the following notification is generated:

```JSON
{"updates":[{"script_name":"dockcheck.sh","installed_version":"vX.Y.Z","latest_version":"v0.7.7",
"release_notes":""xmpp template added, ranges for selection possible, restart-stack and -s reworked.""}]}
```

 Note the invalid double quotes. Expected JSON:

```JSON
{"updates":[{"script_name":"dockcheck.sh","installed_version":"vX.Y.Z","latest_version":"v0.7.7",
"release_notes":"\"xmpp template added, ranges for selection possible, restart-stack and -s reworked.\""}]}
```

Reason is that the escaped quotes are interpreted by `printf` in:
https://github.com/mag37/dockcheck/blob/37052d13772d67f21a7f6e80e266469234a79653/notify_templates/notify_v2.sh#L323-L324

Since I'm only using the file channel (nifty way to do a simple API), I'm solving the issue with the following, but I lack the setup to test it and ensure it wouldn't break other notification channels:
```bash
printf -v MessageBody "%s" "${FormattedOutput}"
```

saschabuehrle:
Fixes #278

`dockcheck_notification()` was assigning `FormattedOutput` with `printf -v MessageBody "${FormattedOutput}"`, which interprets escape sequences and breaks JSON output for file notifications. This switches to `printf -v MessageBody "%s" "${FormattedOutput}"` so escaped quotes are preserved as-is.

Greetings, saschabuehrle

### DIFF
--- a/notify_templates/notify_v2.sh
+++ b/notify_templates/notify_v2.sh
@@ -321,7 +321,7 @@ dockcheck_notification() {
         format_output "dockcheck_update" "$(output_format "${channel}")" "Installed version: <insert_text_iv>\nLatest version: <insert_text_lv>\n\nChangenotes: <insert_text_rn>\n" "$1" "$2" "$3"
 
         # Setting the MessageBody variable here.
-        printf -v MessageBody "${FormattedOutput}"
+        printf -v MessageBody "%s" "${FormattedOutput}"
 
         printf "\nSending dockcheck update notification - ${channel}"
         exec_if_exists_or_fail trigger_${template}_notification "${channel}" || \


mag37:
The issue you're "solving" is an ongoing discussion and solutions need to be tested further. Next time it would be nice if you'd take part in the discussion before contributing a PR with an already proposed (in said discussion) solution.

As this also seems like an AI-agent autoslop (excuse me if it isn't) I'll just close this, as further testing and discussion is required.

PR #280, opened 2026-04-12
Add flag/option to hide the "No Updates availabe" message and only show updateable images

Impact123:
As per title.

### DIFF
--- a/default.config
+++ b/default.config
@@ -22,6 +22,7 @@
 #MaxAsync=10                # Set max asynchronous subprocesses, 1 default, 0 to disable.
 #MonoMode=true              # Monochrome mode, no printf colour codes and hides progress bar.
 #Notify=true                # Inform - send a preconfigured notification.
+#OnlyShowUpdateable=true    # Hides the "No updates available" message and only shows updateable images
 #OnlyLabel=true             # Only update if label is set. See readme.
 #OnlySpecific=true          # Only compose up the specific container, not the whole compose. (useful for master-compose structure).
 #PrintMarkdownURL=true      # Prints custom releasenote urls as markdown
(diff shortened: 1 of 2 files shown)


mag37:
Thank you for the contribution and sorry for not responding earlier. 

I've tested it and it behaves as promised, though before I merge - may I ask the reasoning and demand for the change? :smile: Just want to understand the use cases.

Impact123:
Sure. Like many I used to use [watchtower](https://github.com/nicholas-fedor/watchtower) in [notify/monitor mode](https://watchtower.nickfedor.com/latest/configuration/arguments/?h=monitor#monitor_only). Recently I've been on a mission to reduce unneeeded services/save memory so I replaced it with a periodic call to `dockcheck` via a daily cronjob. I don't really want to get a notification each day that my containers are up to date though. I only care about the outdated ones. This is supposed to help filter for that without having to create some complicated and failure prone regular expressions for `grep` to filter it myself. Think of it as a quiet mode that only reports on _actionable_ things.

mag37:
Very nice and thank you for explaining! I'll merge it :)

I understand the use case and that you might already have other things set up for that workflow - but have you looked into the different ways of running notifications from dockcheck straight away? You can write to a file in different formats for some API-style workflow, or tie it to a messaging service etc. 

Anyway, thank you for contributing this!

Impact123:
Not yet. I might look into that later.

mag37:
Just by changing the config to enable notification to file 
`dockcheck.config`
```sh
NOTIFY_CHANNELS="file"
FILE_PATH="${ScriptWorkDir}/updates_available.txt"
FILE_OUTPUT="json"
```
you'll get a file from each run looking like this:
```sh
{"updates":[{"container_name":"containerABC"},{"container_name":"containerXYZ"}]}
```
Or without the json-format:
```sh
🐋 Containers on HostName with updates available:
containerABC
containerXYZ
```

I'll still merge this PR - but just another alternative depending on workflow!

PR #281, opened 2026-04-15
File notification and JSON output fixes

### LINKED ISSUE #278: Self-update notification produces invalid JSON when using the file channel

vorezal:
* Truncate notify file(s) on first write and then append

* Strip wrapping double quotes from Dockcheck update tempcsv variable before jsonification

### DIFF
--- a/notify_templates/notify_file.sh
+++ b/notify_templates/notify_file.sh
@@ -1,5 +1,5 @@
 ### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
-NOTIFY_FILE_VERSION="v0.1"
+NOTIFY_FILE_VERSION="v0.2"
 #
 # Leave (or place) this file in the "notify_templates" subdirectory within the same directory as the main dockcheck.sh script.
 # If you instead wish make your own modifications, make a copy in the same directory as the main dockcheck.sh script.
@@ -13,10 +13,14 @@ trigger_file_notification() {
     UpperChannel="FILE"
   fi
 
+  FileTruncVar="${UpperChannel}_TRUNC"
   FilePathVar="${UpperChannel}_PATH"
-
   NotifyFile="${!FilePathVar:=${ScriptWorkDir}/updates_available.txt}"
 
-  echo "${MessageBody}" > ${NotifyFile}
+  if [[ ${!FileTruncVar:=0} -eq 0 ]]; then
+    > ${NotifyFile}
+    declare -g ${FileTruncVar}=1
+  fi
 
+  echo "${MessageBody}" >> ${NotifyFile}
 }
--- a/notify_templates/notify_file.sh
+++ b/notify_templates/notify_file.sh
@@ -18,9 +18,9 @@ trigger_file_notification() {
   NotifyFile="${!FilePathVar:=${ScriptWorkDir}/updates_available.txt}"
 
   if [[ ${!FileTruncVar:=0} -eq 0 ]]; then
-    > ${NotifyFile}
+    echo "${MessageBody}" > ${NotifyFile}
     declare -g ${FileTruncVar}=1
+  else
+    echo "${MessageBody}" >> ${NotifyFile}
   fi
-
-  echo "${MessageBody}" >> ${NotifyFile}
 }
--- a/notify_templates/notify_v2.sh
+++ b/notify_templates/notify_v2.sh
@@ -205,7 +205,7 @@ format_output() {
         FormattedOutput=$(jq --compact-output --null-input --arg updates "${tempcsv}" '($updates | split("\\n")) | map(split(",")) | {"updates": map({"script_name": .[0], "installed_version": .[1], "latest_version": .[2]})}')
       elif [[ "${UpdateType}" == "dockcheck_update" ]]; then
         # dockcheck update case
-        FormattedOutput=$(jq --compact-output --null-input --arg updates "${tempcsv}" '($updates | split("\\n")) | map(split(",")) | {"updates": map({"script_name": .[0], "installed_version": .[1], "latest_version": .[2], "release_notes": (.[3:] | join(","))})}')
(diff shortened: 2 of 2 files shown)


vorezal:
Intended to address #278. Nothing incredibly complicated.

- Use a variable flag/switch to identify whether or not to truncate the out file during the first run of the file notification function.
- Strip the wrapping double quotes before sending the Dockcheck update release notes string to JSON. I still want them included in CSV output since that field can have commas.

mag37:
Nice one @vorezal ! Thank you for squeezing in the time. 
My schedule is packed - I'll try to run it through a few tests, see if I can set up a json-specific notification too to properly test it.

vorezal:
I tested json and csv output, though only with the FILE notification type. Others, especially notification types that themselves send json, would be great to verify, but both output types look to be working as expected. I suppose a side effect is that any literal double quotes in Dockcheck release messages will be removed, but I think that would happen on the order of rarely to never, and we'd need to remove or escape them regardless to avoid breaking json formatting anyway.

The switch to appending rather than replacing the content of the notification file does (currently at least) result in multiple json objects in the file rather than a single combined object and it doesn't really identify the type of update for each object directly (they are all just nested under a generic key named "updates" in each object). Those behaviors could almost certainly be changed, but would be more custom cases to handle.

vorezal:
Give me a tiny bit of time before we merge this actually. Instead of truncating separately I think I'd prefer to just do the first operation as an overwrite. I'll make a quick adjustment shortly.

vorezal:
OK that should do it. Probably not really that necessary, but I just didn't like the idea of having a moment where the file was empty.

mag37:
I've not had time to setup any notification that actually processes json in the other end, but I've tried sending json-formatted to both telegram and xmpp with **<CHANNEL>_OUTPUT="json"` which worked just fine and looks like correct json (in plaintext messages).

Very clean solution @vorezal , nice one! Thank you for squeezing it in.

mag37:
I just did some more testing for other reasons, and realized that the file content persist when there's no new updates.

So let's say day1 I get the file-output with 2 containers having available updates. I got some automation to show me that they've got updates.

Then I update those containers and the scheduled check runs again day 2 - but then there's no available updates, though the file still contains the 2 containers from the previous day.

Or am I missing something?

faelnor:
Apologies for the delayed participation, I'm struggling with my time too 🫤 
The escaping issue is fixed, thanks @vorezal! 

> The switch to appending rather than replacing the content of the notification file does (currently at least) result in multiple json objects in the file rather than a single combined object and it doesn't really identify the type of update for each object directly (they are all just nested under a generic key named "updates" in each object). Those behaviors could almost certainly be changed, but would be more custom cases to handle.

AFAIK this is invalid json so I'd prefer if we avoided it. Ideally all updates would go into a single array so I wrote a more direct handling of json output for the file channel in 78478b5e69f8a8497940f32ae140cd56735115ca 

You can look at it and tell me what you think. I kept the FILE_TRUNC check although I can't find a use case where I wouldn't want to start with an empty file (unless we add a timestamp but this brings unnecessary complexity).

> I just did some more testing for other reasons, and realized that the file content persist when there's no new updates.

@mag37 In my case I have a wrapper script that takes care of clearing the output file whenever the script is run (it also formats the output for use by my homepage dashboard and handles a few other checks). It would be nifty to have dockcheck itself take care of clearing the file, I guess the most generic approach would be to have a hook for notification channels to do something at the start of each dockcheck run, but as far as my personal usage goes, I am happy with the current file notification behavior.

vorezal:
Thanks for the contribution! I honestly didn't even think to switch on the output format within notify_file.sh. Maybe I was unconsciously keeping the notification template agnostic of the output format. I'll look at your commit and test it out, but I'm sure we can get this or something very similar in to handle this case one way or another to maintain a single top level element, which it is true is strictly required for a valid json file.

As for leaving the file unchanged when there are no updates, yeah that is mostly intentional. For nearly all notification channels I figure we don't want empty notifications going out every time if there aren't any, though the behavior is perhaps unusual for the file channel specifically. I handle this for specific channels by checking a configuration setting in dockcheck.config that overrides that behavior and does write out an empty notification if there are no updates. My configuration for my file notification channel looks like this:

```
FILE_TEMPLATE=file
FILE_SKIPSNOOZE=true
FILE_ALLOWEMPTY=true
FILE_CONTAINERSONLY=true
FILE_OUTPUT=csv
```

Though of course the `FILE_CONTAINERSONLY=true` setting may not be needed in this case since @faelnor seems interested in Dockcheck updates as well, and FILE_OUTPUT would be `json` but I think the rest is relevant. The output will be `None` in the case of csv output, or `{"updates": []}` for json. If we're certain that the file channel should always allow empty notifications we could change this behavior to make it the default, but then I might need to change some logic to allow people to leave it unchanged instead if they have a use case for that. I can't think of one, but perhaps there is?

Anyway, thanks again for the suggestions and code. Pretty certain we can accommodate.

PR #282, opened 2026-04-24
Update file JSON notification format

vorezal:
* Output a single top level JSON element when writing out file notifications

* Handle case where a Dockcheck update is not available but container updates are

* Simplify logic with variables and a function to manage output text and write mode

### DIFF
--- a/notify_templates/notify_file.sh
+++ b/notify_templates/notify_file.sh
@@ -28,8 +28,8 @@ trigger_file_notification() {
   FilePathVar="${UpperChannel}_PATH"
   NotifyFile="${!FilePathVar:=${ScriptWorkDir}/updates_available.txt}"
 
-  if [[ ${!FileTruncVar:=0} -eq 0 ]]; then
-    declare -g ${FileTruncVar}=1
+  if [[ ${!FileTruncVar:=true} == "true" ]]; then
+    declare -g ${FileTruncVar}="false"
   else
     if jq empty "${NotifyFile}" 2>/dev/null; then
       FileOutput=$(jq --compact-output --argjson msg "${MessageBody}" '.updates += $msg.updates' "${NotifyFile}")
--- a/notify_templates/notify_file.sh
+++ b/notify_templates/notify_file.sh
@@ -31,8 +31,8 @@ trigger_file_notification() {
   if [[ ${!FileTruncVar:=true} == "true" ]]; then
     declare -g ${FileTruncVar}="false"
   else
-    if jq empty "${NotifyFile}" 2>/dev/null; then
-      FileOutput=$(jq --compact-output --argjson msg "${MessageBody}" '.updates += $msg.updates' "${NotifyFile}")
+    if $jqbin empty "${NotifyFile}" 2>/dev/null; then
+      FileOutput=$($jqbin --compact-output --argjson msg "${MessageBody}" '.updates += $msg.updates' "${NotifyFile}")
     else
       if ! grep -xFq "None" ${NotifyFile}; then
         WriteMode="Append"
--- a/notify_templates/notify_file.sh
+++ b/notify_templates/notify_file.sh
@@ -1,10 +1,21 @@
 ### DISCLAIMER: This is a third party addition to dockcheck - best effort testing.
-NOTIFY_FILE_VERSION="v0.2"
+NOTIFY_FILE_VERSION="v0.3"
 #
 # Leave (or place) this file in the "notify_templates" subdirectory within the same directory as the main dockcheck.sh script.
 # If you instead wish make your own modifications, make a copy in the same directory as the main dockcheck.sh script.
 
+write_file_notification() {
+  if [[ "$3" == "Overwrite" ]]; then
+    echo "$1" > "$2"
+  else
+    echo "$1" >> "$2"
+  fi
+}
+
 trigger_file_notification() {
+  local WriteMode="Overwrite"
+  local FileOutput="${MessageBody}"
+
   if [[ -n "$1" ]]; then
     file_channel="$1"
     UpperChannel=$(tr '[:lower:]' '[:upper:]' <<< "$file_channel")
@@ -18,9 +29,16 @@ trigger_file_notification() {
   NotifyFile="${!FilePathVar:=${ScriptWorkDir}/updates_available.txt}"
 
   if [[ ${!FileTruncVar:=0} -eq 0 ]]; then
-    echo "${MessageBody}" > ${NotifyFile}
     declare -g ${FileTruncVar}=1
   else
-    echo "${MessageBody}" >> ${NotifyFile}
+    if jq empty "${NotifyFile}" 2>/dev/null; then
+      FileOutput=$(jq --compact-output --argjson msg "${MessageBody}" '.updates += $msg.updates' "${NotifyFile}")
+    else
+      if ! grep -xFq "None" ${NotifyFile}; then
+        WriteMode="Append"
+      fi
+    fi
   fi
+
+  write_file_notification "${FileOutput}" "${NotifyFile}" "${WriteMode}"
 }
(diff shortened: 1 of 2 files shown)


vorezal:
Intended to address issues raised after #281 was merged. This should output a single top level JSON object for file type notifications. It also fixes an edge case for CSV output when there is no Dockcheck update available which would cause both "None" and container updates to appear in the final file.

I made a couple of design decisions that were different from the original code proposal @faelnor submitted, but still open to thoughts and weighing of pros and cons of this approach vs theirs. Checking if the file was already JSON seemed to keep the notification template code somewhat separate from the logic within notify_v2.sh, and I still like the idea of a single write without a separate truncate step.

notify_templates/notify_file.sh
```diff
@@ -18,9 +29,16 @@ trigger_file_notification() {
   NotifyFile="${!FilePathVar:=${ScriptWorkDir}/updates_available.txt}"
 
   if [[ ${!FileTruncVar:=0} -eq 0 ]]; then
```
faelnor: This will fail on an unbound variable if FILE_TRUNC is set to _true_ or _false_ rather than an integer value. Probably worth updating the check to keep config flags consistent.

faelnor (approved):

faelnor:
Other than the consistency suggestion, I like the changes. Nice way of checking for valid json.

I haven't dug into how it works, but I have done a few tests with FILE_OUTPUT="text" and FILE_ALLOWEMPTY=true and couldn't get the script updates to show up. Am I correct that if FILE_ALLOWEMPTY is set to true, FILE_CONTAINERSONLY is implied? I do not use any of these so I might be trying to use mutually incompatible settings.

vorezal:
Good call. Adjusting the value to true/false is probably a good idea. At first I intended for the variable to be a counter but changed my mind.

Unfortunately I have not been able to reproduce the issue with FILE_OUTPUT="text" or the other variables. None of them are mutually exclusive or imply any kind of change or dependency on one another. Are you running Dockcheck with -i and the -n or -a flags? Dockcheck.sh itself has some logic that it will only send a notification about a new script version being available, which includes writing to this file, if it is in some kind of automatic mode rather than interactive. You'll need to run it with -in or -ia to see an available script update in notifications.

notify_templates/notify_file.sh
```diff
@@ -17,10 +28,17 @@ trigger_file_notification() {
...
   NotifyFile="${!FilePathVar:=${ScriptWorkDir}/updates_available.txt}"
 
-  if [[ ${!FileTruncVar:=0} -eq 0 ]]; then
-    echo "${MessageBody}" > ${NotifyFile}
-    declare -g ${FileTruncVar}=1
+  if [[ ${!FileTruncVar:=true} == "true" ]]; then
+    declare -g ${FileTruncVar}="false"
   else
-    echo "${MessageBody}" >> ${NotifyFile}
+    if jq empty "${NotifyFile}" 2>/dev/null; then
```
mag37: Don't know how we've done elsewhere in **notifications** but in the main script we've got the `jqbin` variable to be able to run a static binary pulled alongside `dockcheck` (and not necessarily in $PATH) for systems not having jq installed nor available through package manager.

mag37:
Good work both of you, thank you for keeping it up! Barely had time to view the discussion this weekend, but looks good!

>Dockcheck.sh itself has some logic that it will only send a notification about a new script version being available

It does, is this someting you guys think we should change? Maybe it's inconsistent - maybe all notifications should be sent either way if the notification option is set.
If you think we should tweak this I'll fix it up in a future patch, to not mess with this scope.

vorezal:
Yeah probably not necessary to mess with the scope now, but I think it would make sense to send it out as a notification if the user has set the -i flag.

notify_templates/notify_file.sh
vorezal: > Don't know how we've done elsewhere in **notifications** but in the main script we've got the `jqbin` variable to be able to run a static binary pulled alongside `dockcheck` (and not necessarily in $PATH) for systems not having jq installed nor available through package manager.

Ah right right. I completely forgot about that. I suppose that's what I get for trying to work quickly. I'll fix that.

vorezal: Apparently I had not thought about this during my work on the notify_v2.sh json formatting at all, but I would definitely consider it a defect. Probably no more changes after this one in the PR, especially since this triggered a version change to notify_v2.sh as well (even if it is minor and a bug fix), but good to have it consistent everywhere. There was one additional script (extras/errorCheck.sh) where jq is referenced directly, but that looks like a standalone script that doesn't do a jq binary check, so it is likely working as intended.

faelnor:
> Unfortunately I have not been able to reproduce the issue with FILE_OUTPUT="text" or the other variables. 

I have not been able to reproduce either, so I agree I must have been using the wrong flags.

> None of them are mutually exclusive or imply any kind of change or dependency on one another.

Even with FILE_TRUNC=false, FILE_ALLOWEMPTY=true will still truncate due to the presence of "None" in the output file, so it feels that those two options shouldn't be used together in most cases unless you know exactly what you are doing.

In any case, I don't think the scope of this PR should be to change this, it's all good for me. Thanks for the changes!

mag37:
> Yeah probably not necessary to mess with the scope now, but I think it would make sense to send it out as a notification if the user has set the -i flag.

Agreed. I'll look into doing that in a cleanup-patch later on.

> In any case, I don't think the scope of this PR should be to change this, it's all good for me. Thanks for the changes!

Depending on how we decide to sort that, it could also be a part of a future cleanup. It's not breaking anything currently and is quite an edge case so should be fine for now. 

Thank you both! Great work. I'll merge this tomorrow if noone finds anything more we should adjust.

mag37:
I'll continue the discussion here for a bit.

>Even with FILE_TRUNC=false, FILE_ALLOWEMPTY=true will still truncate due to the presence of "None" in the output file, so it feels that those two options shouldn't be used together in most cases unless you know exactly what you are doing.

I'm not sure if I grasp the concept of `FILE_TRUNC` correctly - but it's there to keep track of if it's the beginning of a "session" or not, to know if it should start fresh or append. And not an option for the user to set in the .config. Or do I misunderstand?

So what happens if `FILE_ALLOWEMPTY`  is set to true that shouldn't happen?

vorezal:
Yep that's the intention of FILE_TRUNC really. Just to track whether or not that specific file notification has been written out before or if it is being written for the first time during the current script execution. I namespaced it like the other variables because technically someone can create multiple file notifications, but I hadn't really considered the possibility that someone might manually set it to false. I would imagine a user would expect that the file should be appended to forever. The issue with that is I didn't think "None" should appear in the file unless there were actually no updates at all when FILE_ALLOWEMPTY is set, so I have a case where the file is overwritten when the content of the file is "None" and we're making another change to the file. I may solve this by just adding a comment stating that the variable isn't intended to be manually overridden in dockcheck.config for now.

mag37:
Glad I understood it correctly then :) 
You don't need to put more work into it @vorezal - I just wanted to know I grasped the situation.

>I may solve this by just adding a comment stating that the variable isn't intended to be manually overridden in dockcheck.config for now.

I think this is very reasonable, I'll do that when I do some other cleaning no worries.

vorezal:
Thinking about it a bit more, it may make more sense to eliminate the special case and allow appending normally, but include a little extra text clarifying if it refers to Dockcheck script updates or container updates. When there are updates available we write out what they are for, so it would follow the same pattern for a lack of updates as well.

mag37:
Sorry completely missed your comment @vorezal  . 
Well - maybe the truncating (non-appending) flow should be default and we could have a config option to toggle appending? And if so also print like a "header" (for non-json?).

I also think we might be going over the top here :) the file output could be pretty generic so people can plug it or modify it to suit their needs. But a clear base to be easily understood is good.

vorezal:
The fun thing is we actually already do have the truncating flow set as a default and a config option that will allow append only if I change the special case logic around "None" a bit. But I think we're OK for now. I'll consider it as a follow-up though. We are going a little overboard with something meant to be generic as you say. Even a header could just make it more difficult for people to parse.

mag37:
Yeah, lets leave it as is for now and consider how we could approach it in the future. If needed.

PR #286, opened 2026-05-13
Recreate stopped containers without starting

chatziko:
This simple PR makes `-s` recreate stopped containers using a simple `docker compose create`, instead of doing start & stop (starting containers and forcefully stopping might have undesirable side effects).

### DIFF
--- a/dockcheck.sh
+++ b/dockcheck.sh
@@ -52,7 +52,7 @@ Help() {
   echo "-p     Auto-prune dangling images after update. Ignored when -b is used."
   echo "-r     Allow checking/updating images created by 'docker run', containers need to be recreated manually."
   echo "-R     Skip container recreation after pulling images."
-  echo "-s     Include stopped containers, returns to stopped state after recreation."
+  echo "-s     Include stopped containers, recreates without starting."
   echo "-t N   Set a timeout (in seconds) per container for registry checkups, 10 is default."
   echo "-u     Allow automatic self updates - caution as this will pull new code and autorun it."
   echo "-v     Prints current version."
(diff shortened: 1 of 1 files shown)


mag37:
Oh that's very nice! Thank you, I'll need to run some testing and should really merge #271 prior to this - I'll see if I can get them both done asap.

chatziko:
No hurry. I did some light testing (just updated one stopped container, it worked fine), but we should test more.

mag37:
Did run it through a few tests now and seems to work as intended, great find @chatziko !
Will merge as is, just gonna see if I can squeeze #271 too.

mag37:
Okay so I merged #271 to main and things got a bit more complicated to merge :)
And while I was looking at solving that , I considered a situation where both `ForceRestartStacks` and `ContCreateOnly` is used.

```sh
        if [[ "$ContRestartStack" == true ]] || [[ "$ForceRestartStacks" == true ]]; then
          ${DockerBin} ${CompleteConfs} down; ${DockerBin} ${CompleteConfs} ${ContEnvs} up -d || { printf "\n%bDocker error, exiting!%b\n" "$c_red" "$c_reset" ; exit 1; }
        elif [[ "$ContCreateOnly" == true ]]; then
          ${DockerBin} ${CompleteConfs} ${ContEnvs} create ${SpecificContainer} || { printf "\n%bDocker error, exiting!%b\n" "$c_red" "$c_reset" ; exit 1; }
        else
          ${DockerBin} ${CompleteConfs} ${ContEnvs} up -d ${SpecificContainer} || { printf "\n%bDocker error, exiting!%b\n" "$c_red" "$c_reset" ; exit 1; }
        fi
```

Then a stopped container would fall into `[ "$ForceRestartStacks" == true ]]` if that's set, and then gets recreated and started - and left in a running state. 
I've not tested this yet, just from trying to re-arrange the code to fit into the new **main**, so I might be mistaken.

mag37:
I've not forgotten about this :) Still have notes on how I'll try to resolve it. Gonna see if I can get it working asap.

PR #308, opened 2026-08-05
fix: resolve unbound variable error

chrjsorg:
- Fixes Bash <4.4 error when ScriptArgs is empty

Hi, please allow this small fix for my amazing QNAP Bash environment... 😫


mag37:
Ah of course! Good call.
Will merge it to main now but won't push a new release until more changes have landed.

PR #309, opened 2026-08-09
Add FromHost reference to update notification

dseven:
Adds a title line including the FromHost to the message body, consistent with other notifications. Without this, it's impossible tell which host a notification pertains to when using methods that don't utilise MessageTitle, e.g. Telegram.


mag37:
Thank you, good call!

